PR #30, opened 2017-08-05
Add an `Installation` section to usage

quodlibetor:
Because copypasta is the best pasta.


xd009642:
Thanks for this. Merging now :+1:

PR #32, opened 2017-08-07
output: sort output paths

mathstuf:
With a `HashMap`, the output order is undefined. Instead, use a
`BTreeMap` and keep the output order stable.


xd009642:
Thanks for the PR!

Issue #45, opened 2017-10-07
Update outdated dependencies


xd009642:
Below is a list of crates to be updated with their latest versions as of Oct 7th

- [x] cargo   0.18.0 to 0.21.1
- [x] clap 2.24.2 to 2.26.2
- [x] gimli 0.13.0 to 0.14.0
- [x] libc 0.2.23 to 0.2.32
- [x] nix 0.8.1 to 0.9.0
- [x] object 0.3.0 to 0.4.1
- [x] quick-xml 0.8.1 to 0.9.4
- [x] rustc-demangle 0.1.4 to 0.1.5
- [x] serde 1.0.8 to 1.0.15
- [x] serde_json 1.0.2 to 1.0.3

Some of these updates will be harder than others. For example gimli and nix updates I expect to be fairly involved as there have been major changes in both libraries. I'll aim to see if I can update some of these libraries as I work and update this comment with the status.

When updating if you can aim to update usage of the library to make use of new features that make the code simpler that would be great. For example, when updated nix, a lot of ptrace_control.rs can be deleted, and the whole file might be able to be removed!

Anyone who wants to contribute choose a library and feel free to comment here or open a PR and correspond there. If you can make a branch named for the crates you are updating as well that would be helpful from a tracking perspective!

UPDATE: All the updates that require no functional code changes are done! All the others will require a small amount to potentially a lot of code changes.

xd009642:
Closing this issue, thanks @andy128k for your efforts.

PR #50, opened 2017-10-10
Update gimli to 0.14.0

andy128k:
Here is my attempt to help.
Related to https://github.com/xd009642/tarpaulin/issues/45

### DIFF
--- a/src/tracer.rs
+++ b/src/tracer.rs
@@ -1,7 +1,5 @@
 use std::io;
 use std::path::{PathBuf, Path};
-use std::ffi::CString;
-use std::ops::Deref;
 use std::fs::File;
 use std::cmp::{Ordering, PartialEq, Ord};
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


xd009642:
Awesome I've got one small comment I'm just going to add to the code but it's pretty minor and once that's done I'll be happy to merge! Thanks a lot :+1:

src/tracer.rs
```diff
@@ -325,15 +332,17 @@ pub fn generate_tracer_data(project: &Workspace, test: &Path, config: &Config) -
     if let Ok(obj) = OFile::parse(unsafe {file.as_slice() }) {
         let analysis = get_line_analysis(project, config); 
         let data = if obj.is_little_endian() {
-            get_line_addresses::<LittleEndian>(manifest, 
-                                               &obj, 
-                                               &analysis, 
-                                               config.ignore_tests)
+            get_line_addresses(LittleEndian,
```
xd009642: Could something like an if let be used to set an endianness variable and then `get_line_addresses(..).map_err(..)` called at the end once. Just a small thing a part from that I'm happy :smile:

andy128k: It seems it is not possible with this version of gimli. Endianity is used as a generic parameter, and, unfortunately, cannot be boxed.

xd009642: So just to double check this code isn't possible?
```
let endianity = if obj.is_little_endian() {
    LittleEndian
} else {
    BigEndian
}
get_line_addresses(LittleEndian,
                   manifest,
                   &obj,
                   &analysis
                   config.ignore_tests).map_err(..)
'''

andy128k: no, it is not :(

xd009642: I guess it makes sense given how the previous version of gimli worked. But in that case everything's great and I'll merge this now, thanks a ton :+1:

philipc: It's not possible because `LittleEndian` and `BigEndian` are different types. You could consider using `RunTimeEndian` instead. That gives simpler and smaller code but a minor runtime performance loss. I doubt parsing dwarf is a major part of the runtime of tarpaulin though.

xd009642: It probably isn't but I don't think the readability gain is large enough to warrant changing it personally. I'd rather have a bit extra speed

andy128k: @philipc Thanks! I've opened corresponding PR https://github.com/xd009642/tarpaulin/pull/52

PR #53, opened 2017-10-10
update cargo to 0.21.1

andy128k:
I've made verbose unpacking of a tuple because it was changed from `(Package, String, PathBuf)` in 0.18.0 to `(Package, TargetKind, String, PathBuf)` in 0.21.1 and indexes were shifted. So, I think it's better to have names to see what's going on.

Relates to https://github.com/xd009642/tarpaulin/issues/45


xd009642:
Thanks for that, merging in now!

PR #58, opened 2017-10-20
Add a --all flag to manually turn on building a whole workspace

vberger:
This way tarpaulin will behave more consistently with the rest of the cargo subcommands.

The previously added default to `--all` really broke my CI, as I have in my workspace a single package on which tarpaulin should be run : the others cannot be built by tarpaulin with their default feature flags require system libs that are not installed on travis (namely Mesa), forcing me to manually exclude all other packages like this:

https://github.com/Smithay/wayland-rs/blob/ecca2d54340da7afca7b364cb53da61935211de9/.travis.yml#L55

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -64,6 +64,8 @@ pub struct Config {
     pub forward_signals: bool,
     /// Features to include in the target project build
     pub features: Vec<String>,
+    /// Build all packages in the workspace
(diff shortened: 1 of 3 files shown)


xd009642:
Ah sorry about that, that was an unexpected scenario, thanks for the PR though :+1:

PR #59, opened 2017-10-23
Use ptrace_getevent

philipc:
Using `ptrace(PTRACE_GETEVENTMSG, ...)` doesn't work with nix v0.9.0.


xd009642:
Thanks for that, merging in :+1:

PR #60, opened 2017-10-30
Add ( and ) to the list of chars in ignored lines.

vberger:
Without this , lines containing only `)?;` were not ignored.

This removes around 150 lines detected as "uncovered" on wayland-rs.


xd009642:
Awesome, thank you very much :+1:

PR #66, opened 2017-11-30
Fix warnings about unnecessary `mut`

robinst:


robinst:
(The build failed on beta Rust, not sure why.)

xd009642:
Thank you again :smile:

Issue #65, opened 2017-11-28
Docker usage (documentation, maybe publish an image?)


robinst:
Hey. I'm using this crate for CI to get a report for pull requests. But sometimes I want to look at the coverage locally, for example while adding something new or adding tests to cover existing code.

As I was running Mac OS X, I tried using tarpaulin through Docker. I figured out how after some attempts. It would be nice if this was documented somewhere. First, here are the steps currently necessary:

1. Prepare a `Dockerfile`:

```
$ cat Dockerfile
FROM rust

RUN apt-get update && apt-get install -y cmake

RUN cargo install cargo-tarpaulin
```

2. Build the image (takes a while):

```
$ docker build -t cargo-tarpaulin .
```

3. Run cargo-tarpaulin for a project. Needs to be run in the project directory:

```
$ docker run --security-opt seccomp=unconfined -v "$PWD:/volume" -w /volume cargo-tarpaulin cargo tarpaulin --out Xml
```

4. Convert to HTML report to look at results in browser:

```
$ pip install pycobertura && pycobertura show --format html --output coverage.html cobertura.xml
$ open coverage.html
```

### Improvements

There are a couple of things that would make this simpler:

1. The `Dockerfile` could be committed to this repository. Then the first step would be to clone this repository.
2. Even better, the resulting image could be published to [Docker hub](https://hub.docker.com/). That would mean the first two steps would be unnecessary, and much quicker (just download the image instead of building it).
3. If tarpaulin could output HTML, we wouldn't need a separate tool to view the results (seems like this is already being worked on in #55)

And of course, we could add the documentation to the README for all of this. I can open a pull request for that, but if we do any of the steps above first, then the documentation becomes shorter.

What do you think? :)

robinst:
> I suppose the docker image would have to be updated every time a new version of tarpaulin is released?

Yes. I think you can configure Docker Hub to do that automatically, for example when a tag is created. I'll fork this project and add the `Dockerfile` and will experiment with how it works!

Issue #67, opened 2017-11-30
Running `cargo tarpaulin` on clean project doesn't output anything while it's building


robinst:
1. Have a project without a `target` folder, so it hasn't been built yet
2. Run `cargo tarpaulin` (version 0.5.5)

Expected: I get some output telling me what's going on
Actual: I don't get anything for a long time, until it finishes building the project

This is the output:

```
$ cargo tarpaulin
Launching test
...
```

But it took a minute or two before it printed "Launching test". I assume it was building the project, but at first I thought there was something wrong so I killed it and tried again.

(Noticed while looking at #65)

xd009642:
I agree, I usually only judge it by my laptop fans. Just added a fix in (it's not hard tbh) so you'll see this in my next release. Or earlier if you choose to install the latest from master

PR #68, opened 2017-12-01
Add Dockerfile for building on Docker hub

robinst:
This is the first part for #65.

You can see the image here: https://hub.docker.com/r/robinst/tarpaulin/

You can try it out in a repository where you want to calculate coverage using this command:

    docker run --security-opt seccomp=unconfined -v "$PWD:/volume" robinst/tarpaulin

If you want more options, you have to run it like this:

    docker run --security-opt seccomp=unconfined -v "$PWD:/volume" robinst/tarpaulin cargo tarpaulin --ignore-tests --out Xml

If you want to run multiple commands, you have to use a shell like this:

    docker run --security-opt seccomp=unconfined -v "$PWD:/volume" robinst/tarpaulin sh -c "cargo build && cargo tarpaulin"

After merging this (or even before), you can set up an automated build in your account using "Create Automated Build" on Docker hub:

screen shot 2017-12-01 at 13 53 20

You will be able to configure when it should be built automatically, e.g. only build tags or also build master. You probably want to remove automatic building of all branches.

Once that's done I can raise another PR to add the instructions to the README :).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM rust
+
+RUN apt-get update && apt-get install -y cmake
+
+COPY . /opt/tarpaulin/
+
+RUN cd /opt/tarpaulin/ && \
+    cargo install && \
+    rm -rf /opt/tarpaulin/ && \
+    rm -rf /usr/local/cargo/registry/cache/
+
+WORKDIR /volume
+
+CMD cargo build && cargo tarpaulin
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@ COPY . /opt/tarpaulin/
 RUN cd /opt/tarpaulin/ && \
     cargo install && \
     rm -rf /opt/tarpaulin/ && \
-    rm -rf /usr/local/cargo/registry/cache/ && \
-    rm -rf /usr/local/cargo/registry/src/
+    rm -rf /usr/local/cargo/registry/
 
 WORKDIR /volume
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,8 @@ COPY . /opt/tarpaulin/
 RUN cd /opt/tarpaulin/ && \
     cargo install && \
     rm -rf /opt/tarpaulin/ && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apt-get update && \
+    apt-get install -y cmake && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY . /opt/tarpaulin/
+
+RUN cd /opt/tarpaulin/ && \
+    cargo install && \
+    rm -rf /opt/tarpaulin/ && \
+    rm -rf /usr/local/cargo/registry/
```
robinst: These are to make the image a bit smaller. And the registry would be out of date for users of this image anyway, between when we build it and they use it.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /volume
```
robinst: This is where the command is going to be run. The `-v "$PWD:/volume"` argument to `docker run` maps the working directory outside docker to `/volume`.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+CMD cargo build && cargo tarpaulin
```
robinst: I think the `cargo build` is not necessary, but I've added it for now. See #67 for why.

xd009642:
Why ever the beta and nightly builds fail is likely a separate issue entirely. Everything looks good so I'm going to merge this in and set up dockerhub, thank you :+1:

robinst:
I saw that you tried building the tag here: https://hub.docker.com/r/xd009642/tarpaulin/builds/bgvliskm2842x9ygyglbamm/

That's not gonna work until the next tag which will contain the Dockerfile. For now, can you set it up to build master, until we want to release this?

xd009642:
Yeah I've currently got a docker image built called latest which is based on the most recent commit on master https://hub.docker.com/r/xd009642/tarpaulin/builds/bggdjcshcsh6dgp2bhpgbhk/ if you want to start the PR for the updates to the readme/documentation that would be helpful :smile:

PR #71, opened 2017-12-07
README: Add instructions for usage with Docker

### LINKED ISSUE #65: Docker usage (documentation, maybe publish an image?)

robinst:
Fixes #65.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,14 @@ stage and therefore may contain some bugs. A lot of work has been done to get it
 working on some example projects and smaller crates so please report anything
 you find that's wrong. Also, check out our roadmap for planned features.
 
-Tarpaulin only supports x86_64 processors running linux. This is because
+Tarpaulin only supports x86_64 processors running Linux. This is because
 instrumenting breakpoints into executables and tracing their execution requires
 processor and OS specific code. It is a goal when greater stability is reached
 to add wider system support, however this is sufficient to run Tarpaulin on
-popular CI tools like Travis. 
+popular CI tools like Travis.
+
+It can also be run in Docker, which is useful for when you don't use Linux but
+want to run it locally, e.g. during development. See below for how to do that.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


xd009642:
Thank you very much :smile:

robinst:
Cool, thank you! :)

Do you wanna post a link to the instructions on Reddit, or mind if I do? Could be interesting to people, now that they can run it locally.

xd009642:
I don't mind if you do, and some of us have always been running it locally :wink:

PR #96, opened 2018-03-23
bump quick-xml to v0.12.1

tafia:
cargo-tarpaulin is targeting version 0.9.0 only. 

The main changes for you are some bug fixes plus the use of failure instead of error-chain.

PR #97, opened 2018-03-28
Bump dependencies

tafia:
Bump all but cargo dependencies.

For cargo, while I easily made it compile, one test was not passing and I didn't have time to dig through it.

Cancel and replace #96


xd009642:
That's fair enough on cargo it's changes normally require a lot more digging sometimes in the cargo source code, thanks for helping out :+1:

PR #103, opened 2018-04-30
Document possible use of `runtime-macros`

### LINKED ISSUE #102: Code coverage for macros
Do you know if it would be possible to trace code coverage in procedural macros when they're being expanded? I'm currently making a procedural macro crate and would like to use tarpaulin to keep track of code coverage, but it just shows that my test cases are running and nothing else--probably because everything else (i.e. the macro expansions) runs at compile time. I haven't looked into it yet, but I expect that if tarpaulin can trace rustc's execution, it'll be able to report on the code coverage in procedural macros.

It might also work for declarative macros, but I doubt it. Probably worth a try, though.

jeremydavis519:
Fixes #102.

The new [`runtime-macros`](https://crates.io/crates/runtime-macros) crate provides all the functionality needed to get Tarpaulin working with procedural macros. I've added a short section to the readme to point users to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,10 @@ pycobertura show --format html --output coverage.html cobertura.xml
 
 Then open `coverage.html` in your browser.
 
+### Procedural Macros
+
+Normally, Tarpaulin can't report on code coverage within the code for a procedural macro. You'll need to add a test that expands the macro at run time in order to get those stats. The [`runtime-macros` crate](https://crates.io/crates/runtime-macros) was made for this purpose, and its documentation describes how to use it with Tarpaulin.
+
 ## Issues and Contributing
 
 Issues, feature requests and pull requests are always welcome! For a guide on


xd009642:
Thanks for this, and thanks for showing me runtime macros! 

One small issue before I merge `its documentation` should be `it's documentation`, saves someone opening a PR later to fix a single apostrophe, once that's done I'll merge it in :+1:

jeremydavis519:
Actually, no, "its documentation" is correct. Adding the apostrophe there is a common mistake, but "it's documentation" can only mean "it is documentation." The reason is that "it" is a pronoun, rather than a noun, and possessive pronouns never include apostrophes. "Its" is like "your" or "his," and "it's" is like "you're" or "he's."

Sorry to get pedantic like that. It's hard to know how much explanation is enough when talking through a comments system.

xd009642:
My mistake, I was unsure and asked someone I assumed would know better! Accepted and merged, and no problem on the pedantry it's my preferred approach :+1:

PR #122, opened 2018-06-29
Aggregate results

### LINKED ISSUE #92: Feature Request: Show only uncovered lines
Maybe it is possible to add a feature which only shows the uncovered lines of a `tarpaulin --verbose` run.
Basically something like:
 `cargo tarpaulin --verbose | grep "hits: 0"`

It would be even nicer, if the output summarizes adjacent lines without coverage like:
`src/mymod.rs: 20-40 - hits: 0`

RadicalZephyr:
This isn't a production quality implementation, it's currently just a hack for my own personal use.  In particular, there are no tests for the new code, and the output format could use some improvement.  If you're interested in merging this feature in I would be happy to clean things up.

For reference, this is what the new output format looks like for the simple example project: 

```
Coverage Results
src/lib.rs: 6
src/unused.rs: 4-6

src/lib.rs: 5/6
src/unused.rs: 0/3

55.56% coverage, 5/9 lines covered
```

Fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,12 +132,6 @@ We recommend taking the minimal rust .travis.yml, installing the libssl-dev
 dependency tarpaulin has and then running Tarpaulin with the version of 
 rustc you require.
 
-The travis-install shell script will install the latest tagged release built
-on travis to your travis instance and significantly speeds up the travis 
-builds. If you want an earlier version of tarpaulin, look at the script to
-see what you need to implement and adjust accordingly or use 
-`cargo install cargo-tarpaulin` and specify the version in the arguments.
-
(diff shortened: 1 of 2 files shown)


xd009642:
Cool, thanks for this. I've just got two minor changes to the report printout. Could you remove the `## ` before the headings and also add a colon after `Tested/Total Lines` to make it consistent with `Uncovered Lines:`

xd009642:
Awesome, thank you. Also, for future reference set the base branch to merge into to the develop branch. I've set it via the github UI and there's no conflicts so we're all good but it saves potential rebasing etc.

RadicalZephyr:
Ah, good to know. :+1:

PR #135, opened 2018-07-18
Don't print a line for files that have no uncovered lines

RadicalZephyr:
I noticed that files with no uncovered lines are still getting printed in the uncovered lines listing.  Thanks for the new release!


xd009642:
Good catch :+1:

PR #137, opened 2018-07-23
Propagate unreachability in coverage

vberger:
This is an attempt at propagating upward unreachability in processed files. The rationale being that code that can be statically determined to be unreachable cannot be tested, and should not be marked as missed in coverage.

The algorithm for propagating coverage I implemented is the following:

- `unreachable!()` macro invocation is unreachable
- a list of statements is unreachable iff at least one of the statements is unreachable
- a loop is unreachable iff its body is unreachable
- an if block is unreachable iff both its branches are unreachable
  - in particular an if with no else branch is *reachable*
- a match block is unreachable iff all its branches are unreachable
  - in particular, a match with no branch is unreachable, as it must be matching on an uninhabited type like an empty enum
- a function or method with an unreachable body is unreachable itself

This is a conservative implementation with minimal assumptions, I probably missed a few heuristics that could propagate more unreachability.

### DIFF
--- a/src/source_analysis.rs
+++ b/src/source_analysis.rs
@@ -29,6 +29,12 @@ pub trait SourceAnalysisQuery {
     fn should_ignore(&self, path: &Path, l:&usize) -> bool;
 }
 
+#[derive(Copy,Clone,Debug)]
+enum SubResult {
+    Ok,
+    Unreachable
+}
+
 impl SourceAnalysisQuery for HashMap<PathBuf, LineAnalysis> {
 
     fn should_ignore(&self, path: &Path, l:&usize) -> bool {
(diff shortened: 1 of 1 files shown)


xd009642:
Cool I'll have a look over this now. One thought/concern is in instances where this leads to big blocks of code marked as not coverable whether that would be intuitive to users or whether they'd assume this is an issue in tarpaulin

src/source_analysis.rs
```diff
@@ -234,22 +240,35 @@ fn process_items(items: &[Item], ctx: &Context, analysis: &mut LineAnalysis) {
...
+            Item::Macro(ref i) => { visit_macro_call(&i.mac, ctx, analysis); },
             _ =>{}
         } 
     }
 }
 
 
-fn process_statements(stmts: &[Stmt], ctx: &Context, analysis: &mut LineAnalysis) {
+fn process_statements(stmts: &[Stmt], ctx: &Context, analysis: &mut LineAnalysis) -> SubResult {
+    // in a list of statements, if any of them is unreachable, the whole list is
```
xd009642: Can there be a test for this. Just something simple like

```Rust
fn test_unreachable() {
    let x: u32 = foo();
    if x > 5 { 
        bar();
    }
    unreachable!();
}
```

vberger:
Added. It was a good finding: adding this test allowed me to find a bug I did not anticipate, statement macro calls like `unreachable!();` are considered as items, not semicolon-ended-expressions...

xd009642:
Ah I've been bitten by that one before, always forget it's a thing.

This all looks good to me, so merging in :+1:

PR #143, opened 2018-08-30
Cleanup trailing spaces

AnderEnder:
Found a lot of trailing spaces in code:

```sh
find src/ -name '*.rs' -exec grep -e ' $' {} \;
```

 Cleaned up with regex.

P.S. I have not touched the data files, but there are 3 files with the same issue.


xd009642:
Thanks, merged in :+1:

PR #145, opened 2018-09-03
run arbitrary binary for coverage report

### LINKED ISSUE #144: Ability to measure code coverage of any binary
My usecase is to run my program under tarpaulin, run some tests (from a Python process which should not be observed) against it, then kill my process and see how much of its code was used.

I had a quick look at tarpaulin's code and it doesn't seem like there's something fundamentally blocking it. I'll try to make a PR -- any input on the implementation would be apprechiated.

(ref getsentry/semaphore#76)

untitaker:
Fixes #144

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -85,6 +85,8 @@ pub struct Config {
     pub varargs: Vec<String>,
     /// Duration to wait before a timeout occurs
     pub test_timeout: Duration,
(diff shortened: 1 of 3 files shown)


untitaker:
I decided to play around a bit more with my fork until I'm happy with the endresult. It might be the case that we need something entirely different.

Issue #25, opened 2017-08-01
Enable non-default features in tarpaulin build


xd009642:
For a tarpaulin run a user may wish to enable non default features that enable additional functionality and tests for a more accurate view of the coverage. Add a features flag to allow feature forwarding to tarpaulin during the build stage.

PR #154, opened 2018-10-06
Add an option for no default features in target build

aschampion:
While additional features could already be specified, this adds the
ability to disable default features. Related to #25.


xd009642:
Looks good to me, merging now :+1:

PR #164, opened 2018-11-25
Update install instructions

senden9:
Otherwise we got an error because of `publish-lockfile`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Tarpaulin is a command-line program, you install it into your linux development
 environment with cargo install:
 
 ```text
-RUSTFLAGS="--cfg procmacro2_semver_exempt" cargo install cargo-tarpaulin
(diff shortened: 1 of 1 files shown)


xd009642:
It does mention it right near the top in bold `Due to unstable features in syn and issues with no packaging tarpaulin with the Cargo.lock file tarpaulin is now a nightly only crate. Replace all calls to cargo tarpaulin with cargo +nightly tarpaulin` 

However, I have noticed there is a typo `no packaging` instead of `not packaging`. I suppose if the README wasn't clear enough as it is, I'd prefer that snippet to be moved under the installation heading, the typo fixed and maybe change the last sentence to something like ```If you don't run nightly by default, replace all calls to cargo tarpaulin with `cargo +nightly tarpaulin`.```

xd009642:
I'm going to close this as I've addressed the typos and I feel it's clear enough as it's at the top and bold.

Issue #163, opened 2018-11-24
Better error reporting


Cogitri:
I've recently had to debug why tarpaulin would fail on Travis but not locally, which was a pain because all tarpaulin prints if it fails to run is `Error during run` which is... very not so helpful. I'd suggest using `env_logger` instead of plain `println`. If you want to I can take a look at it and open a PR for it later on.

xd009642:
Error reporting is a bit improved from work undertaken for #160 but I'm open to any improvements to the user experience. My only concern is that some of the printouts of the test binaries I can't stop so the formatting may end up looking strange with `env_logger` printouts combined with the test binary printouts

xd009642:
@Cogitri so I've decided moving to `env_logger` is the right call if you're still willing to do the PR? If not I can have a look at it

Cogitri:
I already did some work, sorry for not updating this. I'll post something till this weekend :)

xd009642:
Oh that's alright, I was half expecting to have to do it myself so that's good news :+1:

PR #177, opened 2018-12-08
[WIP] Improve error reporting with 'failure' crate

Cogitri:
Attempts to tackle #163

As the title suggests, this is very much a work in progress.


xd009642:
Tarpaulin uses a develop branch for general development and keeps master just for releases and any readme updates that people need to see. So if you could move this to branch off the develop branch and also update the tests to use the new interface.

I do like how the code has changed in terms of the error enum and the `launch_tarpaulin` function though. Failure definitely makes things look nicer :+1:

Cogitri:
> Tarpaulin uses a develop branch for general development and keeps master just for releases and any readme updates that people need to see.

Oh, oops, right.

> So if you could move this to branch off the develop branch and also update the tests to use the new interface.

Will do once I'm done :)

src/errors/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    TestRuntimeFail(String),
+    /// Failed to parse
+    #[fail(display="Error while parsing: {}", _0)]
+    Parse(std::io::Error),
+    /// Failed to get test coverage
+    #[fail(display="Failed to get test coverage! Error: {}", _0)]
+    TestCoverage(String),
+    #[fail(display="{}", _0)]
+    IO(std::io::Error),
+}
```
Cogitri: Please do take a good look at the error messages here, I guess there might be room for improvements

src/main.rs
```diff
@@ -73,15 +89,9 @@ fn main() -> Result<(), RunError> {
...
-    println!("Tarpaulin finished");
-    res
-    /*match run(&config) {
-        Ok(()) => println!("Tarpaulin finished"),
-        Err(e) => {
-            println!("Error during run");
-            std::process::exit(e);
-        },
-    }*/
+    run(&config).map_err(|e| error!("{}", e.to_string()));
```
Cogitri: This prints errors in a pretty manner:

screenshot from 2018-12-08 18-48-54


However, we still have to switch println!s to `info!` or the likes.

xd009642:
Okay so I went over all the `println!` statements in `src/statemachine.rs` to categorise them and here is a list. I've put the line number based on the latest commit on your branch and also a brief description of the message just in case they've moved around.

src/statemachine.rs

* Line 89 timeout error - `error!`
* Line 102 timeout error - `error!`
* Line 124 existing printout `error!` capture the type and print it out in a `debug!`
* Line 178 unexpected signal `warn!`
* Line 182 error starting test `error!`
* Line 191 failed to trace child threads `error!`
* Line 201-204 PIE error `error!`
* Line 210 address clash `warn!`
* Line 214 failed to instrument `error!`
* Line 319 error during run coverage maybe wrong `warn!`
* Line 320 error message in e `error!`
* Line 384 385 disabling hit count stuff `warn!`

Another thing, if log level debug is set I think it should override the verbose config option to true.

I'll be running this on a few different things and seeing what it's like to see if I have any other comments but good work :+1:. Once my current comments are addressed the minimum after that is to update the README mentioning how to change the log level and also updating the example tarpaulin output in the command line section

Cogitri:
> Another thing, if log level debug is set I think it should override the verbose config option to true.

Uh? We set log level to debug only _if_ verbose is set

> src/statemachine.rs ...

I suppose we should handle the errorengis Result then instead of `error!`

xd009642:
So does the `RUST_LOG` environment variable not potentially affect the behaviour as shown in the [env_logger](https://github.com/sebasmagri/env_logger) readme? I'm happy to just use the additional flags instead of `RUST_LOG` I'm just conscious that if it has an effect on the output people may start to rely on it

src/main.rs
```diff
@@ -22,13 +25,26 @@ fn is_dir(d: String) -> Result<(), String> {
...
 
+fn set_up_logging(verbose: bool) {
+    let mut builder = Builder::new();
+
+    // NOTE: This overwrites RUST_LOG
+    if verbose {
+        builder.filter_module("cargo_tarpaulin", log::LevelFilter::Debug);
+    } else {
+        builder.filter_module("cargo_tarpaulin", log::LevelFilter::Info);
+    }
```
Cogitri: We basically overwrite RUST_LOG here in favour of the verbose switch

Cogitri:
> I'm just conscious that if it has an effect on the output people may start to rely on it

Doing it the way I'm currently doing it makes it impossible to do stuff like:

`RUST_LOG=cargo_tarpaulin=info,someothercrate=debug`

I suppose we could probe for the RUST_LOG env variable and only apply the log filter if it isn't set.

xd009642:
Nah how it is currently looks fine to me, I hadn't checked the changes in `src/main.rs` yet. I'll give it a proper look either today or tomorrow :+1:

xd009642:
> I suppose we should handle the errorengis Result then instead of `error!`

I meant to ask about this, but what did you mean by this? 

Also I'm not 100% on this stylistically

```
[INFO  cargo_tarpaulin] Coverage Results:
[INFO  cargo_tarpaulin] Uncovered Lines:
[INFO  cargo_tarpaulin] src/lib.rs: 6
[INFO  cargo_tarpaulin] src/unused.rs: 4-6
[INFO  cargo_tarpaulin] Tested/Total Lines:
[INFO  cargo_tarpaulin] src/lib.rs: 5/6
[INFO  cargo_tarpaulin] src/unused.rs: 0/3
```

I think I'd prefer something like below for readability.

```
[INFO  cargo_tarpaulin] Coverage Results:
Uncovered Lines:
src/lib.rs: 6
src/unused.rs: 4-6
Tested/Total Lines:
src/lib.rs: 5/6
src/unused.rs: 0/3
```

Also, don't want to be annoying but is it possible to change it to print the log level and tarpaulin instead of cargo_tarpaulin?

So yeah comments in this and the changes to `src/statemachine.rs`. Plus the README still needs to be updated to show how tarpaulin's output has changed if you run it on the simple project and then I think I'll be happy merging this :smile:

Cogitri:
> Also I'm not 100% on this stylistically
> ...
> I think I'd prefer something like below for readability.

Ah, I agree, that does look better.

> Also, don't want to be annoying 

Do _not_ worry about that, "annoy" me as much as you like, that way we get the best result :)

>  is it possible to change it to print the log level and tarpaulin instead of cargo_tarpaulin?

Ah, so something like `[INFO tarpaulin] log_message`?

> I meant to ask about this, but what did you mean by this?

Oops, did some fairly big typos in my comment there :)

Instead of doing `error!` we should use result and do `return Err(RunError::SomeError(ErrorMessage));` - unless we can recover from the error in the function itself? Haven't looked at it yet.

xd009642:
> Ah, so something like `[INFO tarpaulin] log_message`?

That's exactly it

> Instead of doing `error!` we should use result and do `return Err(RunError::SomeError(ErrorMessage));` - unless we can recover from the error in the function itself? Haven't looked at it yet.

Well some like a test timing out means you may not have the full results or even any from that test but the other tests tarpaulin runs might be fine. Whereas ones like the PIE error means everything is going to be messed up. So it's kind of some but bad results versus completely useless. In reality I don't think any of them are recoverable but for some you might want to carry on doing the update loop until the test has exited just so you don't leak resources.

Cogitri:
> In reality I don't think any of them are recoverable but for some you might want to carry on doing the update loop until the test has exited just so you don't leak resources.

Alright. Shouldn't the calls to `data.cleanup()` do that though? I'll think of something nice to call that when we get a StateMachine error.

xd009642:
@Cogitri are you going to have time to work on this, this week or next?

Cogitri:
Will work on this this evening and tomorrow :)

Cogitri:
I'll go ahead and do one-change-per-commit commits so it's easier for you to comment on them. I guess you can squash them once this is good to go.

Cogitri:
Alright, tarpaulin works again, but error handling is a little broken right now due to me being unsure what can be recovered from and what not. Will have to pour more work into this tomorrow.

xd009642:
Yeah I was going to go for `0.7.0` but don't bother changing the version number in Cargo.toml, I'm going to try and slide a few more things in before a release

xd009642:
@Cogitri how's it going? You'll have to do a rebase because I updated some dependencies in the `Cargo.toml`.

Cogitri:
Ah, sorry. I have to write a paper until Friday, will work on it over the weekend

xd009642:
That's alright, it's getting near the Christmas holidays anyway so I don't expect it anytime soon.

Cogitri:
Alright. Note that does work in theory, but I have to figure out why tarpaulin sometimes errors out when it shouldn't - I guess sometimes TestState::Timeout/Abort have been handling gracefully where they aren't now.

Cogitri:
Hm, how can I properly debug tarpaulin? When I run tarpaulin from `./target/debug/cargo-tarpaulin tarpaulin` it seems as if it can't get the path to the cargo executable :/

xd009642:
Have you had a look at the wiki? I've got some developer guidance on there to try and make tarpaulin understandable to newcomers.

If I have time later I'll clone your branch and see if I can see what's wrong.

Cogitri:
Ah, thanks. I've been kind of sparse on time for Rust the last few days.

src/statemachine.rs
```diff
@@ -164,121 +139,107 @@ pub struct LinuxData<'a> {
 
 impl <'a> StateData for LinuxData<'a> {
 
-    fn start(&mut self) -> Option<TestState> {
+    fn start(&mut self) -> Result<TestState, RunError> {
         match waitpid(self.current, Some(WaitPidFlag::WNOHANG)) {
-            Ok(WaitStatus::StillAlive) => None,
+            //TODO: Which TestState is correct here?
+            //Ok(WaitStatus::StillAlive) => Ok(None),
```
xd009642: So this function traditionally returned the new `TestState` if it had changed. So if it hadn't changed I returned `None`. So for `Ok(WaitStatus::StillAlive)` you should return the current test state unaltered. If you want it to return the current TestState not a delta. For `Ok(_)` I agree that should be an error as the current behaviour just had it timeout and then kill the test due to the timeout.

xd009642: Returning the current state again means you will have to pass it into `start` as an argument, or you could mimic what you did for `wait` and return a `Result<Option<TestState>, RunError>`

src/statemachine.rs
```diff
@@ -80,50 +70,37 @@ impl TestState {
...
-    pub fn step<T:StateData>(self, data: &mut T, config: &Config) -> TestState {
+    pub fn step<T:StateData>(self, data: &mut T, config: &Config) -> Result<TestState, RunError> {
         match self {
             TestState::Start{start_time} => {
-                if let Some(s) = data.start() {
-                    s
-                } else if start_time.elapsed() >= config.test_timeout {
-                    println!("Error: Timed out when starting test");
-                    TestState::Timeout
+                data.start()?;
```
xd009642: You need to save this returned state somewhere otherwise the `TestState::Initialise` won't be returned

xd009642:
So I finally had some time to look this over and it doesn't look that far off! You'll need to rebase though as HTML reports are in and they have a few `expect(..)` based error printouts. It's not a big change though.

Cogitri:
Ah, thanks for taling a look! Will work on it tomorrow

Cogitri:
Alright, I've solved the conflicts, please take a 2nd look.

Cogitri:
Also, I'll open a Rust-2018 PR once this is merged.

xd009642:
This all looks good to me. Well done and thanks for the work :+1:

Issue #55, opened 2017-10-12
HTML reports


a-dma:
It would be nice to have the possibility to generate local HTML coverage reports [like lcov does](http://ltp.sourceforge.net/coverage/lcov/output/index.html) without having to rely on external tools like travis or coverall. Mostly, having a percentage indicators and a function breakdown while being able to browse the code would be a very valuable addition. imo.

xd009642:
So that is the main approach I was thinking of for the road-map item of "annotated coverage reports". I realise that's really open ended and non-specific but I was leaving my options open. I haven't really worked on it yet because no one's expressed an interest in the feature. Until now!

I'll start investigating and try to get something in for my 0.6.x release. I'm not gonna make any promises yet just cause it's early days. But if I have any updates I'll let you know here.

a-dma:
Great, thanks a lot.
As always time is scarce, but I'm happy to lend a hand when possible

xd009642:
Thanks for the offer. Firstly, I'll need to choose a crate to let me generate the html. I'm having a look at some options but if you have any specific recommendations I'd be open to suggestions :+1:

a-dma:
There's a ton of template crates (there's a ton of everything Rust right now...). Have a look at https://github.com/rust-unofficial/awesome-rust#template-engine. I think something simple with as few dependencies as possible would be nice here, I think `horrorshow-rs` looks interesting.

Otherwise I would probably use `serde` and implement a custom deserializer into HTML.
Just my two cents of course!

mathstuf:
What about generating markdown and letting that get rendered using a markdown crate?

xd009642:
So there's currently significant process on this that hasn't yet been pushed to github yet. There are more interactive elements like a project browser to let people browse their results. No one get too excited yet but wait and see what's in store :wink:

mathstuf:
OK, sounds good :) .

guidiego:
Hey @xd009642 Im totally interested in this project, could I help in this one?

xd009642:
Hmm I'm not sure how easy this one is to jump into, it's being done as part of a near complete rewriting of the report module. Also the HTML reports are actually designed, it's now just a case of doing the rewrite and filling in the template.

If you're open to other issues, #54 #113 and #63 might be good ones to work on. (Though #113 depends how comfortable you are reading the cargo source code).

I was also going to look into what windows support would take - though that is a big addition

guidiego:
[...]

PR #182, opened 2018-12-20
Initial version of HTML report

andy128k:
Related to https://github.com/xd009642/tarpaulin/issues/55

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1125,6 +1125,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "serde"
 version = "1.0.82"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "serde_derive 1.0.82 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "serde_derive"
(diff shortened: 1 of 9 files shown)


xd009642:
There's already work on this in the `issue_55` branch which is a big overhaul of the whole reporting system to make it easier to add new reports in future. So I'd kind of prefer any work to be a continuation of the existing work.

AnderEnder:
@xd009642,

Looks like `issue_55` branch is unfinished and it does not look like there is a big progress over the time.

Moreover this PR's implementation is much simpler, cleaner and uses more modern web approach. And it is already ready to use.

I would suggest to use this implementation as a base for HTML reports.

xd009642:
@andy128k if you sort the `Cargo.lock` merge conflict I'll merge this PR. 

Issue_55 did add things like source_tree for folder based browsing of the project similar to what you see with coveralls or codecov and syntax highlighting but also was reworking all the report types to make them more testable so was turning into a monster that I cba to finish. Some parts of it are definitely worth taking but I'll work on moving them over in smaller more manageable chunks later on

andy128k:
folders are added

xd009642:
merging now!

PR #183, opened 2018-12-22
Implement `--release` option.

### LINKED ISSUE #181: Release/debug_assertions for tests
**Comparison between kcov and tarpaulin**
tarpaulin: [link](https://codecov.io/gh/daxpedda/unchecked_unwrap/src/b94637633c8c4bd5246375102b06e0b48c29e893/src/lib.rs#L47...78)
kcov: [link](https://codecov.io/gh/daxpedda/unchecked_unwrap/src/c4f7173099c2ec2b882229b9209dd791e80436f2/src/lib.rs#L47...78)

**Problem**
Tarpaulin is right about the uncovered lines because this code only runs in release mode.
I made two test files, one that is supposed to be run in debug and one in release, but tarpaulin doesn't support building tests with `--release`.

**Solution**
- Implement a flag that builds in release mode but retains debug info.
- Implement a flag that disables `debug_assertions`.

**Notes**
- I'm actually not sure if the test results can be combined in a useful way.
- I tried to write a patch for it but didn't succeed (https://github.com/daxpedda/tarpaulin):
  - I tried doing this: `copt.build_config.release = config.release`, which does enable release mode but doesn't disable `debug_assertions`.
  - I also couldn't figure out how to enable debug info at the same time.
  - Woud love to help out and make a PR for it, but I really need some pointers for the Cargo API.

daxpedda:
After some trial and error it finally worked.
Fixes #181.

**Questions**
- Instead of `env::var(rustflags)` we should use `build_config.extra_rustc_args` or `build_config.extra_rustc_env`. But I couldn't figure out how they work, I don't consider important for that PR.
- I added `-C debuginfo=2` to all kind of builds, I assume that is not going to have any unintentional effects?

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -66,6 +66,8 @@ pub struct Config {
     pub varargs: Vec<String>,
     /// Duration to wait before a timeout occurs
     pub test_timeout: Duration,
+    /// Build in release mode
+    pub release: bool,
 }
 
 impl<'a> From<&'a ArgMatches<'a>> for Config {
@@ -95,6 +97,7 @@ impl<'a> From<&'a ArgMatches<'a>> for Config {
             excluded_files:     get_excluded(args),
             varargs:            get_list(args, "args"),
             test_timeout:       get_timeout(args),
+            release:            args.is_present("release"),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


xd009642:
`-C debuginfo=2` shouldn't have any effects unless a user tried to override it. But when tarpaulin relies on debug info to work I would advise against anyone doing that anyway.

This all looks good to me - tried it out myself and code looks good. So I'm going to merge this in, thanks for your work :+1:

PR #187, opened 2018-12-26
Flip around `--skip-clean` and make it `--force-clean`.

daxpedda:
I noticed that changing `RUSTFLAGS` should always result into a complete rebuild from scratch by cargo.
Anything else should be considered a bug by cargo.

**Questions**
- `Adds a clean stage to work around cargo bugs that may affect coverage results` is that fine or is that too much?
- `--force-clean` is fine? Maybe `--clean` or `--add-clean-stage`? (these are horrible suggestions I know)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ output correctly reports the lines the test hits.
 ```text
 cargo tarpaulin -v
 Running Tarpaulin
-Cleaning project
 Building project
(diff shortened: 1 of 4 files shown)


daxpedda:
I was just thinking, instead of removing the flag, I could leave it and mark it as deprecated.
Could either throw and error or actually retain behaviour.

xd009642:
I'm fine with force-clean as a name, merging now :+1:

PR #188, opened 2019-01-02
Upgrade cargo to allow new manifest features

untitaker:
This fixes "unable to get packages from source" errors when a dependency
uses 2018 edition features such as dependency renaming: https://github.com/rust-lang/cargo/issues/5653

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ publish-lockfile = true
 name = "cargo-tarpaulin"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


xd009642:
This change is already on the develop branch so will be in the next release - I only use master for releases so always go off what's on develop for any PRs :smile:. I was hoping to get another two features over the line before my next release but it looks like this is a necessity now!

I'll set up another release of the latest version to remedy this and close this as it's already been implemented.

PR #199, opened 2019-01-08
Rust 2018

Cogitri:


xd009642:
Nice one, I was still building up the courage to move to 2018! Merging now

Issue #165, opened 2018-11-28
Tarpaulin developer wiki feedback


xd009642:
I've started work on a github wiki for tarpaulin https://github.com/xd009642/tarpaulin/wiki/Developers this issue is for any feedback on the content and how helpful it is, missing things, things that need improving and general improvements.

It's still a work in progress but the following checklist shows roughly what's done and pending.

**Checklist**

* [x] Project overview 
* [ ] Cross platform support overview 
* [ ] Building other projects 
    * [ ] Some words on using cargo 
    * [ ] Linker arguments 
* [ ] Source analysis
    * [ ] Code that should be included but may get missed 
    * [ ] Code that shouldn't be in coverage stats but might mistakenly get added
* [ ] Instrumenting code 
    * [ ] Explanation of DWARF
    * [ ] Usage of tools like objdump to debug issues
* [ ] Process tracing 
    * [x] Launching and tracing tests basics
    * [ ] ptrace and it's difficulties 
    * [ ] Issues or difficulties tarpaulin faces 
    * [x] Breakpoint system 
    * [x] State machine
    * [ ] Something about decision/condition coverage 
* [ ] Generating reports - need to assess once #55 is merged in

**Questions**

* How best to advertise wiki?
* What information would be most useful?
* Is the list of files associated with each section helpful?
* What should be the balance of Tarpaulin architecture and design choices versus technology used like ptrace, DWARF, ELF etc.

rep-nop:
![image](https://user-images.githubusercontent.com/24203105/49191976-052f1780-f347-11e8-8b89-9efc7cad387f.png)

Missing a closing paren.


----


```
Tarpaulin also currently only supports x64 systems to AMD and Intel 64 bit processors. This is due to needing processor specific opcodes to add breakpoints and processor registers - the program counter. However, this in theory wouldn't be a large amount of work. But ptrace support can vary wildly between different operating systems and architectures.
```
This paragraph is worded a bit awkwardly -- specifically the first couple sentences.


----


![image](https://user-images.githubusercontent.com/24203105/49192217-217f8400-f348-11e8-800f-3b83b896978f.png)


----


```diff
- When one thread stop they all stop
+ When one thread stops they all stop
```

As well as various lacking punctuation throughout :P


----

Anywho, overall its pretty descriptive and made of mostly well-formed thoughts that make sense :+1: no major complaints on my end.

xd009642:
[...]

PR #206, opened 2019-01-21
Badges update

sesam:
Changing the README only, so I guess this goes on master to make it show up on github.

Adds a link to Development wiki page. Should it be bigger? Inspired by #165 "How best to advertise wiki?"


xd009642:
No that looks noticeable enough! 

As a note I did initially change the base branch to develop instead of master as that's where PRs typically get merged into but then decided to change it back to master so people can see this change without having to wait for the next release. But any future PRs if you go off develop there will be more changes :+1:

PR #211, opened 2019-02-07
properly visit test functions

dgriffen:
Test functions were not properly processed. This meant that flags such as `--ignore-panics` would not be applied leaving lines marked as uncovered in the test functions.

### DIFF
--- a/src/source_analysis.rs
+++ b/src/source_analysis.rs
@@ -416,10 +416,8 @@ fn visit_fn(func: &ItemFn, analysis: &mut LineAnalysis, ctx: &Context) {
     }
     if ignore_span {
         analysis.ignore_span(func.span());
-    } else if test_func {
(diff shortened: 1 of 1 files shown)


xd009642:
Could you fix the compile issues then I can properly review this. I'll also have to have a look at what's already there just to understand the issue a bit further so if you could provide an example where tarpaulin behaves incorrectly that would be appreciated :+1:

dgriffen:
Sure, I didn't realize there was a compile error since I was able to make it work on my machine. [here](https://codecov.io/gh/dgriffen/nestor/src/master/nestor/src/handler.rs) is an example of it marking the `panic!` branches of some of my tests as missed.

xd009642:
`GitHub API: Forbidden` But I'll clone the repo later and run it locally

xd009642:
I've ended up adding this functionality in another commit so closing this PR

PR #222, opened 2019-04-16
Make tests pathing relative to Cargo manifest

TheDan64:
On my system(Ubuntu 16 in WSL), tests were failing because the test directory paths were being built incorrectly. For instance: `tarpaulin/tests/data/ifelse/tests/data/proc_macro/Cargo.toml` was one. Instead, paths should be built relative to the `CARGO_MANIFEST_DIR` envvar which Cargo sets and is the location of your Cargo manifest. As a byproduct, this also lets you run `cargo test` in other directories (ie the test dir).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,10 @@ autotests = false
 [[bin]]
 name = "cargo-tarpaulin"
 
+[[test]]
+name = "integration"
+path = "tests/mod.rs"
+
 [dependencies]
 cargo = "0.34"
 chrono = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ categories = ["development-tools"]
 keywords = ["cargo", "cargo-subcommand", "testing"]
 #publish-lockfile = true
 edition = "2018"
+autotests = false
 
 [[bin]]
 name = "cargo-tarpaulin"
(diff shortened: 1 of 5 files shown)


TheDan64:
Hmm, `mod.rs`, `line_coverage.rs`, and `doc_coverage.rs` seem totally unrelated. If we made the latter two submodules of `mod.rs`, I bet we could share this as a helper function?

xd009642:
Yeah I was meaning to do that at some point, if you make that change (if you haven't already) I'll look at the others :+1:

TheDan64:
Should be all set now, I think

TheDan64:
@xd009642 Any chance you have time to take a look at this?

TheDan64:
There - should be fixed forreal

TheDan64:
Hmm, not sure why that test failed - doesn't seem to be failing locally

xd009642:
I'll have a look today at some point once I get back home. That failure does look like a strange one though :confused:

xd009642:
Well it happens on my machine and I have no inkling why... I'm just trying out some different things now to see if I can resolve it or better understand it

xd009642:
Found it! The environment variable changes are persisted after each test causing the issue. I've just pushed a small change to develop that prevents this as it could cause issues with some users workflows as well

TheDan64:
Should I rebase this branch then?

xd009642:
If there's anything else you want to add then yes, if not I'm happy accepting this and merging it as I've seen it work with my latest change present :+1:

TheDan64:
I'm all set, merge away!

xd009642:
Cool merged :smile:

PR #237, opened 2019-06-16
Fix readme Docker commands for Windows

Ortham:
A very minor quality of life improvement: the readme's Docker commands can almost be copy/pasted into a PowerShell prompt, but `$PWD:` needs to be `${PWD}:`.


xd009642:
Cool thanks for this :+1:

PR #244, opened 2019-08-18
calculate relative paths to base_dir (env::current_dir(), or root arg)

PSeitz:
ignore lines containing "}else{"
add --manifest-path option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ file.
 ## [Unreleased]
 ### Added
 - Created `CHANGELOG.md`
+- Add --manifest-path option
 
 ### Changed
+- Ignore lines containing "}else{"
+- Use relative file paths to base_dir (env::current_dir() or --root option if set)
 
 ### Removed
 
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -2,6 +2,7 @@ pub use self::types::*;
 
 use std::path::{Path, PathBuf};
 use std::time::Duration;
+use std::env;
 
 use clap::ArgMatches;
 use coveralls_api::CiService;
@@ -17,6 +18,8 @@ mod types;
 pub struct Config {
     /// Path to the projects cargo manifest
     pub manifest: PathBuf,
+    /// Path to the projects cargo manifest
+    pub root: Option<String>,
     /// Types of tests for tarpaulin to collect coverage on
     pub run_types: Vec<RunType>,
     /// Flag to also run tests with the ignored attribute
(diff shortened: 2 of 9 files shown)


xd009642:
Cool I'll have to check this over just to make sure I'm happy but a brief skim it looks good. I recently added a changelog to the project could you also update that :+1:

PSeitz:
Sure, I added the changes to the changelog.
I added "Use relative file paths to base_dir" under #Changed, but the behaviour should be the same. There are only differences in the path with the manifest-path option.

xd009642:
So I'm not sure if this works as expected, given a manifest path in a different directory:

```
cargo run -- tarpaulin --manifest-path tests/data/simple_project/Cargo.toml 
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/cargo-tarpaulin tarpaulin --manifest-path tests/data/simple_project/Cargo.toml`
Error: "Failed to parse Cargo.toml! Error: invalid path url `tests/data/simple_project`"
```

Also, it seems like it would be simpler to not add the manifest member into the config struct or remove root as they're basically just the same data but the types are different and one has the `Cargo.toml` appended to the path. And if the project root isn't set but the manifest-dir is this could create errors due to the redundant information

PSeitz:
Whoops, I fixed it. I tested the wrong branch. I think there might be a case where someone wants to use both, have paths calculated relative to root, but still use different manifest-paths. What do you think?

There are some things, which are weird:
Coverage in coveralls is wrong for `rocket_server.rs`. The test run from the root folder ignores most of the lines, which are still ignored when merged:
https://coveralls.io/builds/25207759
When I execute only the tests in the server directory, the coverage looks fine.
https://coveralls.io/builds/25209623

In Codecov, the merge looks fine:
https://codecov.io/gh/PSeitz/veloci/tree/a9f7d6f49f976f2fdc7c205ff1c799d639953d42

PSeitz:
Also, root and manifest-path are not the same, since only root affects the calculation of the base_dir.

xd009642:
> Also, root and manifest-path are not the same, since only root affects the calculation of the base_dir.

So in that case the help text should be amended for `--root` maybe mention that if a separate manifest dir isn't specified it will look for the manifest at the location in `--root`

xd009642:
However, apart from that (and the pending changelog update) it all looks good

PSeitz:
I updated the PR

xd009642:
Awesome, I'll merge it in now :+1:

PR #249, opened 2019-09-04
add --output-dir option

### LINKED ISSUE #245: Support custom report/output directory
First I just want to say thank you for this awesome project! ❤️ 

I was wondering if you'd be receptive to a new option to allow the user to control where the Xml/Html/etc. report files are written?

The current behavior writes the files to the cwd, so if I run `cargo tarpaulin -o Xml -o Html` I end up with:
```bash
./cobertura.xml
./tarpaulin-report.html
```

I'd love to have a new option (perhaps something like `--output-dir` or `--report-dir`) that would let me specify a target directory to emit those reports. For example:
`cargo tarpaulin -o Xml -o Html --output-dir .reports`
would give me:
```bash
./.reports/cobertura.xml
./.reports/tarpaulin-report.html
```

I'd be happy to take a pass at implementing this and submit a PR as well

calebcartwright:
Resolves #245 - Adds a new option `--output-dir` that allows the user to specify the target directory to write the reports to. 

I didn't add any automated tests specifically for this, but please let me know if/where I need to add any!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ file.
 ## [Unreleased]
 ### Added
 - Created `CHANGELOG.md`
-- Add --manifest-path option
+- Add `--manifest-path` option
+- Add `--output-dir` option
 
 ### Changed
 - Ignore lines containing "}else{"
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -42,6 +42,8 @@ pub struct Config {
     pub branch_coverage: bool,
     /// Output files to generate
     pub generate: Vec<OutputFile>,
+    /// Directory to write output files
+    pub output_directory: PathBuf,
     /// Key relating to coveralls service or repo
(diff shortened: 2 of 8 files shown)


xd009642:
Just an initial skim and you should use the output-dir option in coveralls as well. If `--debug` is given the coveralls json is saved as a report. Also, at some point add an entry to the changelog for the new argument :smile:

calebcartwright:
> Just an initial skim and you should use the output-dir option in coveralls as well. If `--debug` is given the coveralls json is saved as a report. Also, at some point add an entry to the changelog for the new argument 😄

Oops 🤦‍♂  I took a look at the coveralls file but didn't look closely enough. Thanks!

calebcartwright:
Please also let me know if I need to squash my commits

xd009642:
No need to squash I just let github handle the merging :+1:

PR #250, opened 2019-09-05
Support JSON output

jmcomets:
While working on a small coverage viewer for Vim, I added basic support for
JSON output. This was previously marked as "not implemented" and would panic.

I didn't do anything fancy, just re-used the existing JSON serializer for
CoverageReport that's written when generating HTML output.

### DIFF
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -43,6 +43,8 @@ pub enum RunError {
     Html(String),
     #[fail(display = "Failed to generate XML report! Error: {}", _0)]
     XML(cobertura::Error),
+    #[fail(display = "Failed to generate JSON report! Error: {}", _0)]
+    Json(String),
     #[fail(display = "Tarpaulin experienced an internal error")]
     Internal,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -345,6 +345,9 @@ pub fn report_coverage(config: &Config, result: &TraceMap) -> Result<(), RunErro
                 OutputFile::Html => {
                     report::html::export(result, config)?;
                 }
+                OutputFile::Json => {
+                    report::json::export(result, config)?;
+                }
                 _ => {
                     return Err(RunError::OutFormat(
                         "Format currently unsupported".to_string(),
--- a/src/report/json.rs
+++ b/src/report/json.rs
@@ -21,7 +21,7 @@ struct CoverageReport {
     pub files: Vec<SourceFile>,
 }
 
-pub fn export(coverage_data: &TraceMap, _config: &Config) -> Result<(), RunError> {
+pub fn export(coverage_data: &TraceMap, config: &Config) -> Result<(), RunError> {
     let mut report = CoverageReport { files: Vec::new() };
     for (path, traces) in coverage_data.iter() {
         let content = match fs::read_to_string(path) {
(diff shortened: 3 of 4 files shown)


xd009642:
So could you move the data into a different representation just to make a nicer report? 

So looking at this exported report turn it from something like:
```
{
  "files": [
    {
      "path": ["/",  "home", "xd009642", "code", "rust", "json", "tests",  "data", "simple_project", "src", "lib.rs"],
      "content": "pub mod unused;\n\n\npub fn branch_test_one(x: i32) -> i32 {\n    if x > 5 {\n        10\n    } else {\n        5\n    }\n}\n\n\n\n\n\n\n#[cfg(test)]\nmod tests {\n    use branch_test_one;\n    #[test]\n    fn bad_test() {\n        branch_test_one(2);\n    }\n}\n",
      "traces": [
        {
          "line": 4,
          "address": 4212944,
          "length": 1,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 5,
          "address": 4212948,
          "length": 1,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 6,
          "address": 4212964,
          "length": 1,
          "stats": {
            "Line": 0
          }
        },
        {
          "line": 8,
          "address": 4212954,
          "length": 1,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 21,
          "address": 4212048,
          "length": 1,
          "stats": {
            "Line": 2
          }
        }
      ],
      "covered": 4,
      "coverable": 5
    },
    ...
```
Into:
```
{
  "files": [
    {
      "path": [
        "/home/xd009642/code/rust/json/tests/data/simple_project/src/lib.rs"
      ],
      "traces": [
        {
          "line": 4,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 5,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 6,
          "stats": {
            "Line": 0
          }
        },
        {
          "line": 8,
          "stats": {
            "Line": 1
          }
        },
        {
          "line": 21,
          "stats": {
            "Line": 2
          }
        }
      ],
      "covered": 4,
      "coverable": 5
    },
  ...
```

xd009642:
Also there's now a PR for an `--output-dir` option that's been merged into develop that should be used to find where to save the json :+1:

jmcomets:
I took into account the `--output-dir` option in the configuration.

For the JSON report's schema, I do agree that it could be cleaned up a bit. I'm a bit concerned that it won't be aligned with the JSON used in the HTML report. I was considering refactoring this to use the same code, but didn't feel like doing it in my first PR.

To change the JSON schema, I mainly need to customize the serialization of the `SourceFile` struct.
Some thoughts:
- The `path` member is currently a `Vec<String>` but does feel like it could be a `PathBuf` or `String`. I can't see why it was split into its components in the first place. 🤷‍♂ 
- I think the `content` member should be included in the final report, mainly to tell if the current source is different from that used to generate the coverage data.
- I believe the `traces` shouldn't include the `address` member, as it doesn't make much sense from a report perspective. Since the struct is shared, this means implementing a custom struct/serialization to exclude it.
- No particular thoughts on the `covered` & `coverable` members. I don't know the exact use-case for them.

xd009642:
So I didn't write the HTML reporting module that was someone elses PR. I imagine the `Vec<String>` would have been to make the file browser easier in the HTML report. and the `covered` and `coverable` are likely to make calculating the percentage easier. And yeah keep the content then it does make it easier.

If you don't fancy getting the HTML report to use the json from the json report just create the custom struct/serialisation. Changing the `Vec<String>` stuff sounds like a bit of a pain and they can always diverge

xd009642:
Closing this as it was provided by #362

PR #254, opened 2019-09-21
Use HashSet for xml package deduplication

jmagnuson:
`Vec::dedup` only removes consecutive elements, but HashSet implicitly dedups regardless of order.

I was running into an issue where codecov was failing to show any files at the root `src/` path. Digging deeper, it turned out that the XML contained multiple `<package name="src" ...>` entries which codecov didn't like. The fix properly generates unique names for the `packages` array.

### DIFF
--- a/src/report/cobertura.rs
+++ b/src/report/cobertura.rs
@@ -38,6 +38,7 @@
 ///   </packages>
 /// </coverage>
 /// ```
+use std::collections::HashSet;
 use std::error;
 use std::fmt;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


rye:
This is a great change! I was wondering why I was missing the files at the root of my `src/` directory, and this makes sense. I was slightly concerned about sorting, but that doesn't seem to be an issue. Thanks for finding and fixing this!

xd009642:
Nice catch :+1: merging now :)

PR #261, opened 2019-10-02
Ignore empty executables when processing doctests

tmiasko:
When a doctest is annotated with `no_run` the persisted executable is an
empty file and tarpaulin would fail to parse it. Ignore empty files
to avoid the issue.


xd009642:
@tmiasko can you apply some fixes so this works in stable as well please? And also update the CHANGELOG once that's done :+1:

tmiasko:
PTAL

xd009642:
Tested it and it works great :+1: merging now then releasing a new version

PR #265, opened 2019-10-07
Add tests for break in loop/for/while loops

camchenry:
This adds some tests for covering the `break` expression in `for`, `while` and `loop` blocks/expressions (per #54). This is my first contribution to this project, so please let me know what I can change to increase the code quality.

Issue #54, opened 2017-10-11
Needs more tests


xd009642:
So tarpaulin is in need of some more tests. Currently I'm relying on a small number of tests and doing informal runs of projects to test things but as more features get added on this becomes harder to do. 

There are a few areas I'd like to tackle, if anyone chooses to help I'm willing to help mentor you through any issues and generally help out.

**Integration tests**
* Test tarpaulin on run throughs of more of the test projects
* Assert against lines that should be hit
* Try with more variations of config options
* Create examples that represent coverage requirements derived from #63 

**Unit tests**
This one is a bit trickier based on the reliance of ptrace or processing a binary in tarpaulin. 
* [x] More unit tests for the source_analysis module would be useful.
* [ ] Look to test report generation - particularly cobertura reports
* [ ] Look for more testable areas that don't rely on external files or ptrace

**Test projects for all Rust expression types**
List taken from [syn](https://docs.rs/syn/0.15.11/syn/enum.Expr.html)
* [x] Box
* [x] Array
* [x] Call - adequately covered by other tests
* [x] MethodCall
* [ ] Tuple
* [ ] Binary
* [ ] Unary
* [x] Lit - covered by any constants in other tests
* [ ] Cast
* [x] Type
* [x] Let
* [x] If
* [x] While
* [x] ForLoop
* [x] Loop
* [x] Match
* [ ] Closure
* [ ] Unsafe
* [x] Block
* [x] Assign - `<expr> = <expr>` form so way more to test than I've done but covered by a lot of others
* [x] AssignOp
* [x] Field
* [ ]  Index
* [ ] Range
* [x] Path
* [x] Reference
* [x] Break
* [x] Continue
* [x] Return
* [ ] Macro
* [x] Struct
* [ ] Repeat
* [ ] Paren
* [ ] Group
* [ ] Try
* [ ] Async
* [ ] TryBlock
* [ ] Yield

**Bonus challenge**
Related to the issue with unit tests. This is potentially a challenging issue and might be ultimately more trouble than it's worth but is at least worth an investigation. It might even result in the genesis of a design that enables future non-Linux based development as I imagine it will result in a coverage tracing trait of some form. 
* Look into stubbing solutions to test tarpaulin with a simulated ptrace

xd009642:
[...]

PR #267, opened 2019-10-09
Add continue expression tests

camchenry:
These changes are very similar to my last pull request (#265), but this time it's for `continue` expressions in loops. I tried to test most of the possible scenarios, but feel free to suggest any.

Completes the `continue` expression item for #54.


xd009642:
:+1:

PR #273, opened 2019-10-23
Format code using 'cargo fmt'

Atul9:
I ran 'cargo fmt' to format the code

### DIFF
--- a/src/breakpoint.rs
+++ b/src/breakpoint.rs
@@ -70,7 +70,11 @@ impl Breakpoint {
     }
 
     /// Processes the breakpoint. This steps over the breakpoint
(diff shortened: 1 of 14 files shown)


rye:
@xd009642 would it be helpful to add `cargo fmt` (and perhaps clippy as well?) as a CI check, à la `cargo fmt --check`, which would exit with a bad status code if the code wasn't formatted? Probably only on stable/beta to prevent nightly from breaking things.

I'd be happy to PR that into the Travis configuration, if so.

xd009642:
Yeah that would be helpful, although if you look on the develop branch you'll see there's been a switch to GitHub actions. Also the old Travis config used to have allow failures on nightly but the new born actions don't.

PR #274, opened 2019-10-24
allow relative paths for manifest-path

PSeitz:
This adds support for relative paths for manifest-path, currently only absolute paths are working.


xd009642:
Yeah on nightly I let some ignored tests run that can fail just to check if things have changed in the compiler and they're sorted. This is fine so merging away

PR #276, opened 2019-10-24
actions: Check formatting before continuing with the build

rye:
This PR changes the Actions configuration to include setup for rustfmt and usage of rustfmt before continuing with the rest of the build. Ideally, poorly-formatted code should trigger the earliest feedback, so `cargo fmt` is run before the rest of the build, on all three versions of Rust.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,6 +21,7 @@ jobs:
         with:
           toolchain: ${{ matrix.version }}
           override: true
+          components: rustfmt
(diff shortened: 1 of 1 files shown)


xd009642:
instead of `cargo fmt --check` it should be `cargo fmt -- --check`

rye:
Oh, oops! Just a moment.

xd009642:
Yeah all looks good, bar the formatting issues that need fixing but they can be done in another pr

Issue #247, opened 2019-09-03
Compatibility with cargo test flags


JP-Ellis:
It would be good if tarpaulin could achieve parity with the `cargo test` flags.  In particular the following come to mind:

- `--lib`
- `--bin <NAME>` and `--bins`
- `--example <NAME>` and `--example`
- `--test <NAME>` and `--tests`
- `--bench <NAME>` and `--benches`
- `--doc`
- and most importantly: `--features <NAME>`

though there are many other flags that `cargo test` has but `tarpaulin` does not.

* [x] `--quiet` - this is the default behaviour flag probably not needed
* [x] `--lib`
* [x] `--bin <NAME> ...`
* [x] `--bins`
* [x] `--example <NAME>...`
* [x] `--examples`
* [x] `--test <NAME>...`
* [x] `--tests `
* [x] `--bench <NAME>...`
* [x] `--benches`
* [x] `--all-targets`
* [x] `--doc`
* [x] `--no-run`
* [x] `--no-fail-fast` (have a feeling this is a default for coverage)
* [x] `-p` `--package <SPEC>...` (--packages in tarpaulin though rename)
* [x] `--all` (did it pre-deprecation)
* [x] `--workspace` (need to add as an alias)
* [x] `--exclude`
* [x] `--release`
* [x] `--profile`
* [x] `--features`
* [x] `--all-features`
* [x] `--no-default-features`
* [x] `--target` (no point only support x64 linux currently and need to run it)
* [x] `--target-dir`
* [x] `--manifest-path`
* [x] `--verbose`
* [x] `--color ` not implementing impacts aesthetics of tarpaulins printouts 
* [x] `--frozen`
* [x] `--locked`
* [x] `--offline`
* [x] `-Z`
* [x] `-h` `--help`
* [x] `-- ARGS ...`

xd009642:
You can pass arguments to the test executable by doing `-- <test args>` I don't know if that resolves any of those flags or not. But I'll go through year and add flags accordingly this weekend

JP-Ellis:
Oh, I misread the documentation then.  I thought that `cargo tarpaulin -- <args>` was used in the same way as `cargo test -- <args>` whereby the arguments go to the test binary.  Still, it would be nice to generally have most flags from `cargo test` be supported by `cargo tarpaulin` so that one can easily run tests and then run code coverage.

xd009642:
[...]

PR #288, opened 2019-11-27
options: add --frozen and --locked flags

mathstuf:
See #247

---
In CI, separating the build and test steps (e.g., to ensure the build doesn't have access to privileged Docker containers to use `ptrace`), the test should take `--frozen` to ensure that the build is up-to-date rather than possibly building the crate again.


xd009642:
Awesome thank you :+1: for next time set develop branch as the base just in case there are newer changes than release

Issue #287, opened 2019-11-24
Support lcov.info format


jwir3:
I know that HTML reports have landed in #182, and the issue related to that mentions lcov. One thing that would be nice is to support lcov-style output files (e.g. lcov.info), so that tools like https://atom.io/packages/lcov-info would work out of the box.

xd009642:
Yeah that sounds like a reasonable feature to add, and probably isn't that hard to do a PR for... I'll try to get to it myself but I'm rather busy at the moment so I'll also mentor anyone who happens across this drawn by the help-wanted label :smile:

xd009642:
Some description of the file format https://github.com/mitchhentges/lcov-rs/wiki/File-format

bojan88:
I have most of this implemented, and will be creating a PR soon

xd009642:
This is now in the develop branch thanks to @bojan88 I'll probably drop a new release this week

PR #320, opened 2020-01-13
Support for lcov.info

bojan88:
* Added lcov.info support (#287)
* Added function names to traces

### DIFF
--- a/src/config/types.rs
+++ b/src/config/types.rs
@@ -22,6 +22,7 @@ arg_enum! {
         Stdout,
         Xml,
         Html,
+        Info,
     }
 }
 
--- a/src/config/types.rs
+++ b/src/config/types.rs
@@ -22,7 +22,7 @@ arg_enum! {
         Stdout,
         Xml,
         Html,
-        Info,
+        Lcov,
     }
 }
 
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -43,8 +43,8 @@ pub enum RunError {
     Html(String),
     #[fail(display = "Failed to generate XML report! Error: {}", _0)]
     XML(cobertura::Error),
-    #[fail(display = "Failed to generate Info report! Error: {}", _0)]
-    Info(String),
+    #[fail(display = "Failed to generate Lcov report! Error: {}", _0)]
+    Lcov(String),
     #[fail(display = "Tarpaulin experienced an internal error")]
     Internal,
 }
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -43,6 +43,8 @@ pub enum RunError {
     Html(String),
     #[fail(display = "Failed to generate XML report! Error: {}", _0)]
     XML(cobertura::Error),
+    #[fail(display = "Failed to generate Info report! Error: {}", _0)]
+    Info(String),
     #[fail(display = "Tarpaulin experienced an internal error")]
     Internal,
 }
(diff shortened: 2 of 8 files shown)


src/report/info.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        writeln!(file, "LF:{}", da.len())?;
+        writeln!(
+            file,
+            "LH:{}",
+            da.iter().filter(|(_, hits)| *hits != 0).count()
+        )?;
+
+        // TODO: add support for branching
```
bojan88: I understand branching information is still not supported in tarpaulin, right?
That's why I left the comment without implementation

src/test_loader.rs
```diff
@@ -128,7 +140,7 @@ where
 fn get_addresses_from_program<R, Offset>(
     prog: IncompleteLineProgram<R>,
     debug_strs: &DebugStr<R>,
-    entries: &[(u64, LineType)],
+    entries: &Vec<(u64, LineType, &Option<String>)>,
```
bojan88: I had to wrestle with type checker, and therefore this change.
Let me know please if this and other changes bellow don't look ok.

src/report/info.rs
xd009642: That's correct

xd009642:
Awesome I'll have a look at this after work 👍

src/config/types.rs
```diff
@@ -22,6 +22,7 @@ arg_enum! {
         Stdout,
         Xml,
         Html,
+        Info,
```
xd009642: If lcov only has one report format can you rename this to `Lcov` otherwise change it to something like `LcovInfo` (I'd prefer `Lcov` if it's clear enough)

src/errors/mod.rs
```diff
@@ -43,6 +43,8 @@ pub enum RunError {
     Html(String),
     #[fail(display = "Failed to generate XML report! Error: {}", _0)]
     XML(cobertura::Error),
+    #[fail(display = "Failed to generate Info report! Error: {}", _0)]
```
xd009642: lcov info report or lcov report, and rename the enum field accordingly

src/report/mod.rs
```diff
@@ -10,6 +10,7 @@ use std::io::BufReader;
 pub mod cobertura;
 pub mod coveralls;
 pub mod html;
+pub mod info;
```
xd009642: Also rename the module to be more lcov specific

src/test_loader.rs
```diff
@@ -55,6 +55,8 @@ pub struct TracerData {
     pub address: Option<u64>,
     /// Length of the instruction
     pub length: u64,
+    /// Funcrion name
```
xd009642: nitpick: typo `Funcrion` should be `Function`

bojan88:
Btw, I tried it with Atom an that coverage plugin described in the issue.
I also tried it with [this extension](https://github.com/ryanluker/vscode-coverage-gutters) in VS Code

xd009642:
Okay I've just tried it with the HTML viewer in the lcov repo and it works great :+1: just make the naming changes suggested and update the `CHANGELOG.md` to mention that support for lcov.info report generation has been added and then I'll merge the PR :smile:

bojan88:
Thanks for the review. I think I addressed all requests.
I hope it's fine now, but don't hesitate to request more changes

xd009642:
Yeah that's all good thanks 👍

PR #341, opened 2020-02-08
Test project for Box expression type

shnewto:
:wave: Hi there! I love this project, thanks so much for all the work that's gone into it :)

I'm interested in taking a swing at contributing a bit and the tests in issue #54 seemed like an appropriate place to start. This PR adds tests for the box expression type.

I wasn't sure whether there are expectations for an expression type's tests beyond generally code that uses it (i.e. any preference on how it's used). I just tried to juggle boxing in a few simple, and hopefully interesting enough ways, using the existing arrays module test as a basic reference for how to go about things.

I'm totally open to feedback and am willing to revise anything you'd like to see changed/added/or other. I'd like to be able to pick off a few more of the tests you listed so anything we can do here to get me heading in the right direction will help!


xd009642:
Generally the methodology is to try and get enough examples of the feature in use that there's a relatively high level of confidence it works.

I did originally intend this for box_syntax and box_pattern features as well, but their RFC has been unapproved so they're probably going to disappear at some point and this looks good to me so merging :+1:

PR #348, opened 2020-02-17
Add match expr test and run in test suite

shnewto:
This commit adds an extra scenario to the suite of match expression
tests. It was added because the changes in commit b8b427cdde6f996df20d4dfbf95a4b19d24a76da
add support for it, so now seems like a good time to account for it :smile: 

This commit also adds the matches tests to those run from
`tests/mod.rs`. I'm not actually sure why it was omitted from that file
though, so maybe that's the wrong thing to do :grimacing:

Thanks for taking a look :raised_hands:

### DIFF
--- a/tests/data/matches/src/lib.rs
+++ b/tests/data/matches/src/lib.rs
@@ -1,3 +1,14 @@
+pub enum LibType {
+    A(usize),
+    B(usize),
+}
+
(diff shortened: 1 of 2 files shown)


xd009642:
It was probably omitted cause it didn't pass : grimace:

Issue #77, opened 2018-01-14
fails on travis with ASLR disable failed: EPERM


kdy1:
```
0.42s$ bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh)
cargo-tarpaulin
after_success.2
178.59s$ cargo tarpaulin --out Xml &&
    bash <(curl -s https://codecov.io/bash)
    
Building project
Launching test
ASLR disable failed: EPERM: Operation not permitted
thread 'main' panicked at 'Failed to trace: Sys(EPERM)', /checkout/src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
Unexpected grab
Error while collecting coverage. EPERM: Operation not permitted
Coverage Results
#
# Omitted 
#
0.00% coverage, 0/302 lines covered
Tarpaulin finished
```

Travis log: https://travis-ci.org/swc-project/swc/builds/328637807#L2967

I use nightly

kdy1:
Closing because it seems like a issue of travis. 

https://github.com/travis-ci/travis-ci/issues/9061

philipc:
Possible workaround is running with `sudo: required` so it's not using containers. I think tarpaulin could avoid needing to disable ASLR with some effort (something like https://github.com/gimli-rs/findshlibs might help).

kdy1:
Oh.. that worked. Thank you.

xd009642:
@philipc thanks for the help, I'll make an issue for looking into using findshlibs to improve tarpaulin. Any input you might have would be appreciated :smile:  

I've also added sudo:required to the example travis.yml in the README so new users don't hit this problem as well.

rfdonnelly:
@xd009642, I setup cargo-tarpaulin using the README on crates.io and ran into this issue since the crates.io README doesn't yet have this change.  Please consider bumping the version to update the crates.io README.

BTW, I discovered cargo-tarpaulin from your recent reddit post.  I was looking for a coverage solution and was turned off by the complex setup of kcov.  Cargo-tarpaulin is so much slicker!

xd009642:
Thanks for the compliment :smile: I did consider releasing a patch so crates.io is up-to-date but latest stable is out Feb 4th iirc and I'm hoping to actually have some more features to merge in. And that way I don't have to do another release to get rid of the warning about stable Rust!

Marwes:
[...]

PR #361, opened 2020-03-01
Clarify needed settings for pre-built binaries.

### LINKED ISSUE #359: How to use the pre-built binaries with travis?
I was hoping to use cargo tarpaulin on travis, in the standard linux configuration which is travis-ci-opal-xenial-1553530491-f909ac5. The readme says:

> Alternative, there is the travis-install shell script will install the latest tagged release built on travis to your travis instance and significantly speeds up the travis builds. You can install via that script using bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh).

It never finds `libssl.so.1.1`. I tried installing the packages `libssl-dev` and `openssl`, but it won't find the dynamic lib...

najamelan:
Closes #359

I think it would be great if tarpaulin had CI itself to verify that all
recommended installation procedures actually lead to working setups.

Unfortunately I don't have the time to set that up for you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,13 +231,18 @@ rustc you require. Tarpaulin is installed in `before_cache` to allow it to be ca
 and prevent having to reinstall every Travis run. You can also replace `cargo test`
 with a verbose run of tarpaulin to see the test results as well as coverage output.
 
+Tarpaulin is ran after success as there are still some unstable features which could
+cause coverage runs to fail. If you don't rely on any of these features you can
+alternatively replace `cargo test` with a call to `cargo tarpaulin`.
+
 For codecov.io you'll need to export CODECOV_TOKEN are instructions on this in
 the settings of your codecov project.
 
 ```yml
 language: rust
-sudo: required
-dist: trusty
+sudo: required # required for some configurations
+# tarpaulin has only been tested on bionic and trusty other distros may have issues
+dist: bionic 
 addons:
     apt:
         packages:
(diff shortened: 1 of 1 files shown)


xd009642:
So I think this can be more succinct if it doesn't mention libssl-dev, after-all that is in the example travis.yml so users should always add it. 

Also I realised why the prebuilt doesn't work on trusty anymore and it's because I switched to github actions and ubuntu-latest on GA is bionic. So maybe adapt the note to say that the prebuilt binaries are built on ubuntu-latest with github actions so the ubuntu distribution for travis should match if using them which at the current time is bionic

najamelan:
Ok, but then the example `.travis.yml` also has:
- `dist: trusty`
- `sudo: required` -> not needed in my tests at least with bionic
- `after_success` -> which means if the code coverage step fails, your ci result will still be green, making problems go undetected?
- `cargo install cargo-tarpaulin` -> so when the text below the example says `alternatively` it's not quite clear what to include for the pre-built binary.
- it doesn't have `pkg-config`, `cmake`, `zlib1g-dev` in the package list. So do you need these to compile from source or not?

What do you want to do about all of these issues? In any case, "allow edits from maintainers is on", don't hold back, this is your project after all.

xd009642:
So `dist: trusty` could either delete or replace with `dist: bionic` or add a note saying it's known to work with both versions. I remember there was an issue with one of the travis ubuntu versions where it didn't work but that was so long ago it might not be an issue.

sudo required seemed to be required by a lot of people, but maybe that depends on which ubuntu version: see https://github.com/xd009642/tarpaulin/issues/77 for reference 

after_success was added because some people complained if their tests and build pass tarpaulin failing to run on the project shouldn't fail CI as it's more a tarpaulin issue if it can't build/run a project

the alternatively bit it was intended that only the install command should have to change. And I'm sure with travis it preinstalls pkg-config, cmake, zlib1g-dev hence why the example config just has libssl-dev.

I'll have a think about what I think is best probably need to personally take some time and figure out how much of the readme needs updating based on the current state of things

xd009642:
I'll try to get past some initial issues just to merge this PR tonight though as some things may require more time to decide on. And obviously if you have any opinions or thoughts I'm always open to suggestions :+1:

najamelan:
Sounds great. If the readme is clear, you will save people a lot of time debugging travis builds, because the feedback cycle is very long... (modify config -> seeing result on travis).

I have still been debugging travis build all day, because switching to bionic breaks wasm-pack...

xd009642:
hmm maybe try using cargo install to install tarpaulin then, or the prebuilt docker image might work better and be a quicker CI time for you?

xd009642:
Also I made some tweaks to the readme, let me know if these improve things

xd009642:
Ah that's likely a hangover from when tarapulin was tied to the specific version of nightly that built it so caching meant a segfault whenever the nightly updated. Can remove it for crates that aren't relying on nightly features that may cause things to go wrong. I'll just do that now

xd009642:
Okay I'm gonna merge this now

PR #362, opened 2020-03-04
Add json support

ethanabrooks:
This PR is based on pull request https://github.com/xd009642/tarpaulin/pull/250. This should provide the same functionality as that pull request, except that I refactored that code in order to expose methods that allowed programmatic access of JSON report contents. The goal is to be able to invoke these methods from an external library in order to access the `CoverageReport` struct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ file.
 
 ### Removed
 
+## [0.11.1] 2020-03-04
+### Added
+- Add support for JSON output, including public functions for querying reports programmatically.
+
+### Changed
+- Pulled `trace` function out of `run` in `main.rs` in order to expose public function for creating
+  `TraceMap` structs.
+
+
 ## [0.11.0] 2020-02-26
 ### Added
 - Change in coverage between last 2 runs in HTML report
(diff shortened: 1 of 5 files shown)


xd009642:
So one of my concerns with #250 was that the internal types may expose too much data to be useable to a user, and may end up with users relying on internal implementation details or representations for whatever they use the outputted json for. It's something I should have probably thought about for the html generation but eh :man_shrugging: 

I suppose being able to serialise the whole thing could be useful for some features I was considering recently so I'll just ignore my inner musings on the json.

Could you just update the changelog and then I'll approve and merge

ethanabrooks:
For sure. I'm actually working right now to minimize the exposure of the inner details -- ideally the user would only be able to access things that would be printed to the output json file. Once I've got that worked out, I'll update the changelog.

xd009642:
So you can use serde attributes like `#[serde(skip_serializing)]` on struct fields

xd009642:
just let me know if you have any questions or when it's ready for final review :smile:

ethanabrooks:
Ok just updated the CHANGELOG. In order to get this to work with my repository, I also had to make a minor change to `src/lib.rs`. I basically pulled a `trace` function out of the `run` function in order to generate `TraceMap` structs. Hope that's ok.

xd009642:
Yeah this looks fine to me, I'll merge it now

PR #367, opened 2020-03-06
Reduce the size of docker builds

pksunkara:
Using builder pattern reduced the size from 1.7 GB to 1.2 GB
Using rust-slim reduced the size from 1.2 GB to 0.5 GB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM rust
+FROM rust:slim as builder
 
 RUN apt-get update && \
-    apt-get install -y cmake && \
+    apt-get install -y libssl-dev pkg-config cmake zlib1g-dev && \
     rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt/tarpaulin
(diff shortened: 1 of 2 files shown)


xd009642:
Why remove the `rm -rf /var/lib/apt/lists/*` surely that reduces the size of that stage in the cached stages with no negative side-effects?

pksunkara:
There are 2 images here, We won't be caring about the size of the first image anymore because it is being used only to build the main image when releasing. The second image is the only one being distributed.

xd009642:
first image takes up disk space on the building pc though, typically in places I work they add no-cost optimisations like that to avoid unnecessary waste

pksunkara:
> first image takes up disk space on the building pc though,

Yeah, it does. But it's docker hub building it, so I didn't mind. I will add them back.

xd009642:
Cool LGTM

Stargateur:
This introduce a breaking change in the docker image

pksunkara:
Not everyone who uses this needs the full rust image. If they want something installed, they can have customer Dockerfile which extends this.

The main reason for me doing this is people who has slow internet. It will take them a full day to download the previous image.

xd009642:
I think a better solution is to mimic the rust docker images and have separate slim docker files. I've reverted this in develop but I'll add in the slim images today probably

PR #384, opened 2020-03-29
html: support coverage diffs with deleted sources

### LINKED ISSUE #382: Html diff fails if a source file disappears
If a source file disappears between two Html report generations, the attempt to find the now-deleted files fails and causes the new report to not be created.

mathstuf:
When a previous result mentions a non-existent file, just ignore the
error; it has probably just been deleted since the last run (different
branch, reworked source tree, etc.).

Fixes #382


xd009642:
Looks good to me just update the changelog section under unreleased and I'll merge it :+1:

PR #405, opened 2020-04-27
Added full toolchain name to create_command

### LINKED ISSUE #404: create_command doesn't follow rustup toolchain specification
https://github.com/xd009642/tarpaulin/blob/02e744e715d2bc2cd327261b03e162ddfb34aa56/src/cargo.rs#L142

The create_command function adds +nightly or +beta to the cargo command. However, if a project is using a toolchain from a locked date through the `rust-toolchain` file, then this command will try and use the wrong toolchain.

For example, on our project, we build a Docker image for CI that installs a specific version of nightly. When we run Tarpaulin we get:

`error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed`

Although `rustup toolchain list` would show that we do have a nightly installed:

`nightly-2020-04-17-x86_64-unknown-linux-gnu`

The full toolchain spec can be found here: https://github.com/rust-lang/rustup#toolchain-specification

AngelOnFira:
This closes #404 

This fixes a problem where toolchains that are locked by date will be overridden by Tarpaulin.

I'm not entirely sure if there are any edge cases this won't handle in terms of toolchains, but the notation `cargo +<toolchain> command` should allow the full toolchain name just fine.

Please let me know if anything needs to be changed :smiley:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ file.
 
 ### Changed
 - Find target folder from metadata if not provided and place reports there (fixes running from packages inside workspaces)
+- Using date-locked toolchains no longer defaults to trying to use a toolchain with the channel name and no date
 
 ### Removed
 
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -145,11 +145,7 @@ fn create_command(manifest_path: &str, config: &Config, ty: &RunType) -> Command
         test_cmd.args(&["+nightly", "test"]);
     } else {
         if let Ok(toolchain) = env::var("RUSTUP_TOOLCHAIN") {
-            if toolchain.starts_with("nightly") {
-                test_cmd.arg("+nightly");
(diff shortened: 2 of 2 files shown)


xd009642:
Could you also add something to handle doctest coverage? Just see if the toolchain starts with nightly and use it, otherwise just add `+nightly` doctest retention is a nightly feature in rust so can't completely avoid the need to add nightly

AngelOnFira:
For sure, I'll take a look into that now.

AngelOnFira:
@xd009642 I believe I added what is needed, but I might have missed something or implemented it in a weird way. I'm not amazing with Rust, and got some help from our dev team. Any feedback is appreciated :smiley:

xd009642:
Yeah that's great thanks, if you just add something to the changelog I'll merge the PR :smile:

AngelOnFira:
Done and done :smiley:

xd009642:
Thanks :+1:

Issue #414, opened 2020-05-01
Feature request: CLI arguments override config files


tonyfinn:
It would be nice if I could override the config file on a per job run with a CLI argument. This is currently special cased for a small number of options, such as [`debug` and `verbose`](https://github.com/xd009642/tarpaulin/blob/master/src/config/mod.rs#L356), so these options provided on the CLI will override those specified in the config file, but other CLI options are ignored when a config file is present.

This would allow using the majority of settings from a config file while only overriding one or two on a per run basis. This would also bring tarpaulin in line with how inbuilt cargo commands treat config options, such as [`cargo build`](https://doc.rust-lang.org/cargo/reference/config.html#build)

The use case where this would be useful for me is to do with a project of mine where I have some [tarpaulin configuration](https://gitlab.com/tonyfinn/olympia/-/blob/master/.tarpaulin.toml) to run different sets of tests, as one set of tests must be run single threaded. I have this committed to the repo so that it also takes effect when local runs of tarpaulin are performed. 

In CI, I would like to also publish results to coveralls. Since I'm building on Gitlab CI, not Travis, I need to use a repo token, which I would like to not distribute in my repo, and instead provide it at runtime on the command line, [as seen in the test script here](https://gitlab.com/tonyfinn/olympia/-/blob/master/build/test.sh).

If you're ok with adjusting CLI options to work this way, I can submit a PR to make the change, it basically boils down to merging all CLI options, not just debug and verbose.

xd009642:
So I had a certain number of CLI options override or extend the options from the config file based on what people would want to extend. If someone wants to extend someone and they can't via the options and it makes sense I'll happily accept the PR :smile:

tonyfinn:
I've created a pull request to add this support to a few of the options I could see myself using, including the coveralls option I was looking for here: https://github.com/xd009642/tarpaulin/pull/419

I did not tackle any of the vec items, like the list of excluded files, packages, features, etc. The precendent set by the exclude option would be for options specified on the CLI to be additive to the options in the config file, but this then provides no way to exclude a normally included feature, or include a normally excluded file.

I can see three options for adding the ability to customize these multiple choice options in the CLI when a config file is present

1. Special case file inclusions to have an exclude and include option where the include option overrides the exclude option. This would allow something like `cargo tarpaulin --exclude tests/*.rs --include tests/test_utils.rs` or the same via mixed CLI/config file options. All other options,  being new, could then have CLI options replace existing config file options.
[...]

PR #419, opened 2020-05-04
Allow more CLI options to override config file.

tonyfinn:
This PR allows more config items to be specified on the command line and override an existing config file on an adhoc basis to allow for per run or per environment overrides.

In particular, it adds support for overriding the following options:

* Target directory for Rust builds, for example if a CI build needs to output to a specific cached location.
* Output directory for Tarpaulin output, for the same reason
* Coveralls related config items (coveralls, ci_tool and report_uri). Primarily the coveralls flag was desired, to allow use in environments other than Travis which require access to a secret API key that you may not wish to include in repo, but I included all three options for consistency.
* Crate root, to allow for a run to target only a single crate in a workspace.

Previous discussion: #414

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ file.
 ### Changed
 - Find target folder from metadata if not provided and place reports there (fixes running from packages inside workspaces)
 - Using date-locked toolchains no longer defaults to trying to use a toolchain with the channel name and no date
+- The following CLI options now take effect even when a custom config file is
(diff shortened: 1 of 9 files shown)


tonyfinn:
Sorry about all the extra pushes here, there was a test issue as the order of configs from the toml parsing function was not guaranteed, then I updated the commit messages to fit in github's UI better.

xd009642:
Looks good to me, just pop an entry in the changelog and we're good to go :smile:

tonyfinn:
Changelog has been updated.

xd009642:
:+1:

Issue #383, opened 2020-03-29
"competition" with IDE integration for checks


mathstuf:
I'm using ALE for Vim which ends up running `cargo check` in the background. Whenever this happens, `cargo tarpaulin` ends up having to rebuild the entire dep tree. Is there some way to make these not conflict so much?

xd009642:
I guess it would be because the rustflags variable differs for tarpaulin runs, so if you get ALE to use the same rustflags that should fix it.

You might need to set up that rustflags so it's just for ALE I'm not sure if rustc/cargo are smart enough to spot variables duplicated in the rust flags or just see the string is different and reruns because tarpaulin can result in things being there twice! 

When I'm on my pc I'll post up the rustflags, otherwise on master they should just be in src/lib.rs

colvin:
I'm having this same issue. `cargo check`, `cargo build`, `cargo test` all conflict with tarpaulin, and require full rebuilds when switching between any of the former and tarpaulin.

I'm also using ALE under Vim, and this causes long periods of delay while the background `cargo check` builds the entire dependency tree. Not ideal.

In the meantime I've used a workaround to create a new target directory, with the caveat that I have to create that directory manually and still have to do a `cargo build` before running tarpaulin if I've done a `cargo clean`.
```
cargo build
mkdir -p target/tarpaulin
cargo tarpaulin --target-dir target/tarpaulin ...
```

This puts the build artifacts in `target/tarpaulin/debug`.

I'd love a better way to do this. Setting `RUSTFLAGS` in the ALE configuration is ok, but the issue likely affects other workflows as well (since manual `cargo check` or `cargo test` is also affected). Perhaps tarpaulin could just use its own target directory? It already creates `target/tarpaulin` to store the `coverage.json` file.

mathstuf:
> cargo tarpaulin --target-dir target/tarpaulin

Is this something that can be set up in `tarpaulin.toml`? Though a better default would be appreciated.

colvinwellborn:
I had no luck with that. The build would fail, although I didn't really dig into why.

xd009642:
Ah so `--target-dir` doesn't work if the directory doesn't exist so I'll add a fix for that. And every option that can be set via the command line args can be set via the `tarpaulin.toml` with the same name as the arg name.

As for changing the default target-dir I'm not 100% sold on that, for people who do run locally it could result in a lot more hard drive space used. Also, as long as it's easily configured I'm not sure how much gain this gives most users.

[...]

PR #422, opened 2020-05-10
config: support CARGO_TARPAULIN_TARGET_DIR as a source for the target dir

mathstuf:
Fixes: #383 

---
This allows users to change the default target directory for their own environment without having to add `tarpaulin.toml` files to every repository changing it for every developer.


xd009642:
LGTM thanks :+1:

PR #423, opened 2020-05-10
replace deprecated failure with Display trait implementation

AnderEnder:
`failure` crate is marked as deprecated: https://github.com/rust-lang-nursery/failure/pull/347

This change removes the `failure` usage and replace with `fmt::Display` trait implementation.


xd009642:
looks good to me, thanks :+1:

PR #428, opened 2020-05-11
Added `Error` implementation for `RunError`

Razican:
This makes `RunError` implement the `Error` trait, for easier interoperation with other libraries, if needed. I also upgraded dependencies and cleaned up `Cargo.toml` a bit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,9 +105,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.50"
+version = "1.0.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


AnderEnder:
@Razican, Is it needed? tarpaulin is not a library and currently error trait is not used.

Razican:
> @Razican, Is it needed? tarpaulin is not a library and currently error trait is not used.

As you wish, I thought it would make sense to implement it in case tarpaulin wants to show a backtrace of the error too :)

xd009642:
hmm I think I'll hold back on this, currently the display trait is enough and the most useful backtraces are probably from the actual test binaries in which case setting the `RUST_BACKTRACE` variable is really the only way to see them.

PR #436, opened 2020-05-12
chore: Favor std Infallible over Void

lzutao:
This might be a breaking change because of changing Err item in FromStr.


xd009642:
It's a slight breaking change but I don't think anyone is using tarpaulin in such a way that it matters. And it is definitely an improvement to the code so thanks!

PR #440, opened 2020-05-14
Bugfix/lookup report from config target dir

azriel91:
Heya, when:

* running `cargo tarpaulin` in `<repo>/member_crate`
* setting `target-dir` to the workspace root: `--target-dir ..`

`cargo tarpaulin` still looks for the previous coverage report to compare with in `<repo>/member_crate/target/tarpaulin`.

This PR fixes that by calculating the report directory from the `target-dir`.


xd009642:
oh thanks for catching that :sweat_smile:

PR #441, opened 2020-05-15
Feature/take in compilation target

azriel91:
Heya, some projects compile to different compilation targets, and so need a way to wire that through to `cargo`. This PR adds that functionality, so users can now go:

```bash
cargo tarpaulin --target x86_64-linux-gnu
```

or in `tarpaulin.toml`:

```toml
[linux_32]
target = "x86-linux-gnu"

[linux_64]
target = "x86_64-linux-gnu"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ file.
 
 ## [Unreleased]
 ### Added
+- Compilation target is now accepted through the `--target` parameter.
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


xd009642:
So tarpaulin is only expected to work on x86_64 Linux, I'm not against this as a change just wondering how practically useful it is?

azriel91:
oh, tarpaulin still runs on x86_64 linux, this is for the compilation target, i.e. if you're cross compiling for different destinations, and would like to test different code paths based on that.

So for example if building for a 32 bit system, you want to take a different code path (perhaps it's more optimal).

xd009642:
Ah okay then that sounds fine, and the code looks good too. If you update the changelog I'll merge it in :+1:

azriel91:
:v: cool thanks!

PR #449, opened 2020-05-24
Improvement/fail on compilation fail

azriel91:
* Swaps `HashMap` for `IndexMap` to retain order of feature configurations in `tarpaulin.toml` when running tests. This makes it easier to see which configuration's testing fails.
* Changes failure log level from `info!` to `error!` so it's easier to see what went wrong.
* Sets exit code to 1 when failing to compile tests.

Screenshot from 2020-05-24 19-33-01

Previously it would silently succeed, and it's hard to tell what failed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,8 +30,7 @@ mod ptrace_control;
 
 pub fn trace(configs: &[Config]) -> Result<TraceMap, RunError> {
     let mut tracemap = TraceMap::new();
-    let mut ret = 0i32;
-    let mut failure = Ok(());
+    let mut tarpaulin_result = Ok(());
 
     for config in configs.iter() {
         if config.name == "report" {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -50,7 +50,8 @@ pub fn trace(configs: &[Config]) -> Result<TraceMap, RunError> {
                 ret |= r;
             }
             Err(e) => {
-                info!("Failure {}", e);
+                error!("{}", e);
+                ret |= 1;
                 if failure.is_ok() {
```
xd009642, line 55: So I'm guessing from the fact failure is only really set and not used there was probably an intent to do something like `failure?;` after the loop just to return an error if one of the tests failed so the error type propagates up. I think that's probably the best approach

azriel91: I've propagated the result up; I decided to keep passing the return code up from the test execution to the `trace` method, and log it -- if it reaches that `Ok(..)` trace result with a non-zero exit code, something strange is going on (if the execution fails it would be `Err(..)`)

xd009642: So I just tested on a project with a test that failed and the behaviour changed from returning a non-zero error code to zero. If a test fails then tarpaulin should still return a non-zero return code to match the behaviour of `cargo test` - partially because other people just use tarpaulin instead of cargo test to decrease the time to run their CI.

I thought I had a test that covers this but it looks like I should add one (but I can do that myself). If you move back to keeping the tracking of the return code and then end it with something like:

```
if ret == 0 {
    tarpaulin_result.map(|_| tracemap)
} else {
    Err(RunError::TestFailed)
}
```

That'll work fine

PR #470, opened 2020-06-03
docs(readme): add [ ] to the config file example

S-Coyle:
I believe that the tarpaulin.toml config file example in the README should include [ ] around the features

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,14 +392,14 @@ the projects manifest or in the root directory that will be used unless
 
 ```toml
 [feature_a_coverage]
-features = "feature_a"
+features = ["feature_a"]
 
 [feature_a_and_b_coverage]
-features = "feature_a feature_b"
+features = ["feature_a feature_b"]
 release = true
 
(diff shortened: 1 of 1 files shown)


xd009642:
Firstly thanks for opening a PR and I'm sorry but I'm going to have to decline this because the example is correct as per how tarpaulin works and also consistent with how cargo test works. I've typed up an explanation below:

If you take the configs project in `tests/data/configs` which uses the features field as a test bed for this. If you replace the list with a string you get:

```
[ERROR tarpaulin] Invalid config file invalid type: sequence, expected a string for key `coverage.features` at line 4 column 1
```

You can also use that project to try out the cargo feature behaviour `cargo test --features feature1 feature2` won't activate feature2 but `cargo test --features "feature1 feature2"` will so the type is a string

S-Coyle:
No problem at all, appreciate the explanation & the work you put into tarpaulin! 🙇‍♂️

PR #499, opened 2020-07-17
Don't report coverage when not running tests

OskarPersson:
When using `--no-run` no coverage should be reported since no tests has been executed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,16 +90,18 @@ pub fn run(configs: &[Config]) -> Result<(), RunError> {
         }
     }
     if configs.len() == 1 {
-        report_coverage(&configs[0], &tracemap)?;
+        if !configs[0].no_run {
+            report_coverage(&configs[0], &tracemap)?;
+        }
     } else if !configs.is_empty() {
         let mut reported = false;
         for c in configs.iter() {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -90,16 +90,18 @@ pub fn run(configs: &[Config]) -> Result<(), RunError> {
...
         let mut reported = false;
         for c in configs.iter() {
-            if c.name == "report" {
+            if !c.no_run && c.name == "report" {
                 reported = true;
                 report_coverage(c, &tracemap)?;
             }
         }
-        if !reported {
+        if configs[0].no_run && !reported {
```
xd009642, line 104: Maybe `configs.iter().any(|x| !x.no_run)` here, or `!configs[0].no_run` at least.

xd009642:
Didn't think of this, thanks! One small comment but no other issues, if you address that and also update the changelog we're good to go :+1:

src/lib.rs
OskarPersson: Fixed, went with the latter one :)

OskarPersson:
Fixed and updated changelog 👍

xd009642:
Awesome thanks for this :tada:

PR #522, opened 2020-08-05
Some Improvements

wcampbell0x2a:
Rather small changes that don't change functionality but rather made the code more readable.

All changes found by using clippy.


xd009642:
Looks good just needs a rust fmt 😁

PR #535, opened 2020-08-18
Removing -C link-dead-code (big change)

xd009642:
* Add more logical lines! This simplifies some things
* Add source analysis of async/await/groups and tuples
* Try to identify any line that can be covered and add it to cover


xd009642:
There's some fundamental issues that stop this from working regarding how the rust compiler can reduce expressions that are essentially no-op i.e. `loop { break; }` that makes it infeasible to identify all coverable lines without reimplementing a lot of compiler internals

PR #547, opened 2020-08-26
Feature/no dead code good bits

xd009642:


xd009642:
bit odd CI didn't run on this when I pushed it so need to PR it :thinking:

PR #548, opened 2020-08-26
Try cargo rustc invocation 

xd009642:


xd009642:
Note to self: This didn't work because cargo rustc is only applied to package not whole workspace. See https://github.com/rust-lang/cargo/issues/8651

PR #562, opened 2020-09-11
add feature for using vendored openssl

sameer:
Add a feature for using the `vendored-openssl` feature in `git2`. This will allow `openssl-sys` to compile OpenSSL from source: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/Cargo.toml#L15

It is useful for installing tarpaulin on environments that don't have OpenSSL.


xd009642:
Thanks, merging 👍

Issue #545, opened 2020-08-24
Add tracing 


xd009642:
Add tracing to tarpaulin instead of logs, span information could be useful when looking at how things evolve over time and could open the gates to things like opentelemetry tracing which can solve issues like my naive rendering for minitarp events often not being able to handle the sheer volume of traces.

Also part of this would involve figuring out if all my log messages are useful and conducive to solving issues

xd009642:
Also eyre could make reporting errors look nicer than it currently does :eyes:

xd009642:
So the initial part of this is done in #572 next step is using spans to make things nicer 👀

PR #572, opened 2020-09-29
move to tracing crate

xMAC94x:
- replace `use log::{..}` with `use tracing::{..}`
- replace the env_logger with tracing-subscriber
- keep most of the behavior (variables debug and verbose) which set the default tracing mode.
- CHANGE BEHAVIOR: RUST_LOG now can overwrite debug and verbose, you can also use it to overwrite it only partially
(i wasn't able to run all the tests locally)
(maybe we need to `tracing-log` crate, in order to capture dependencies logs)

First steps of #545 . It still needs a check if the current logging is good.

Also note:
- I had serious problems running tarpaulins tests. So i can't say if it works perfectly.
- removed a useless `use std::io::Write;`

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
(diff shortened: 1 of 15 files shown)


xd009642:
Awesome, thanks. I'll have a proper look at this after work but from an initial glance it looks okay

xMAC94x:
Hey you're welcome. I want to reduce stress to you and might find some other locations where we can enhance tracing (or using spans). This could be done in a follow up, sure. take your time (:

xd009642:
Yeah this all looks good, and enhancing tracing etc can be done in a follow up. Definitely feels better than trying to do it all in one big go 😁

PR #578, opened 2020-10-01
528 - rough version of fix.

ruari-almost:


src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+    None
+
+    // TODO handle [target.thumbv7em-none-eabihf] triplet flags etc.
+}
+
+fn build_config_path(base: &Path) -> PathBuf {
+    let mut config_path = PathBuf::new();
+    config_path.push(base);
+    config_path.push(".cargo");
+    config_path.push("config.toml");
```
xd009642, line 466: so the file may be toml format but it's just called `.cargo/config` so just remove the toml extension

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
     value
 }
 
+fn look_for_rustflags_in(path: &Path) -> Option<String> {
+    if let Ok(contents) = read_to_string(path) {
+        let value = contents.parse::<Value>().unwrap();
```
xd009642, line 439: You can use the try operator here to return early with a `None` value by doing: `let value = contents.parse::<Value>()?;`

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+
+        if let Some(build_value) = value.get("build") {
+            let build_table = build_value.as_table().unwrap();
```
xd009642, line 442: Likewise another place to use `?`

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+
+            if let Some(rustflags) = build_table.get("rustflags") {
+                let vec_of_flags: Vec<String> = rustflags
+                    .as_array()
+                    .unwrap()
```
xd009642, line 447: And another `?` location :smile:

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+                    .into_iter()
+                    .map(|x| String::from(x.as_str().unwrap()))
```
xd009642, line 449: And one final unwrap to take care of

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+            }
+        }
+    }
+
+    None
+
+    // TODO handle [target.thumbv7em-none-eabihf] triplet flags etc.
+}
+
+fn build_config_path(base: &Path) -> PathBuf {
```
xd009642, line 462: Can change base's type into `impl AsRef<Path>` then create the mut config path with `let mut config_path = PathBuf::from(base.as_ref());`

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+}
+
+fn gather_config_rust_flags(config: &Config) -> String {
+    if let Some(rustflags) = look_for_rustflags_in(&build_config_path(&config.root())) {
+        return rustflags;
+    }
+
+    if let Ok(cargo_home_config) = env::var("CARGO_HOME") {
+        if let Some(rustflags) =
+            look_for_rustflags_in(&build_config_path(&PathBuf::from(cargo_home_config)))
```
xd009642, line 478: If you make the change suggested with the `impl AsRef<Path>` above you can just pass in `cargo_home_config` without constructing the temporary `PathBuf`

src/cargo.rs
```diff
@@ -432,6 +434,56 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+        {
+            return rustflags;
+        }
+    }
+
+    return "".to_string();
```
xd009642, line 484: Remove the explicit return at the end of the function. You could just end this line with `String::new()` which would look more idiomatic

xd009642 (commented):
A few comments on different parts. Nothing major though :+1:

src/cargo.rs
ruari-almost: I tried changing this but the compiler got very cross with me. 

It seemed to be because parse() returns an Result<F, F::Err> as opposed to an Option?

Is there an idiomatic way round this?

xd009642: ah you can use `.ok()` to turn a result into an option first so `let value = contents.parse::<Value>().ok()?;`

src/cargo.rs
ruari-almost: Do you mean inside the map()?

xd009642: Yeah, I guess there might be an `as_str_lossy` but I'm not sure what the type is.

ruari-almost: I couldn't find a lot of info on as_str_lossy()? 

If I add a ?, the compiler says I dont have the correct return type:

                     .map(|x| String::from(x.as_str()?))
    |                          -----------------^^^^^^^^^^^-
    |                          |                |
    |                          |                cannot use the `?` operator in a closure that returns `std::string::String`
    |                          this function should return `Result` or `Option` to accept `?`

which kind of makes sense I guess but is there a something sneaky I can do here?

xd009642: ah okay. I looked at the toml docs properly.

So `.as_array()?` to remove that unwrap. And then for the Map you can do `.filter_map(|x| x.as_str()).map(|x| x.to_string())`

filter_map takes in a function that returns an option and discards the `None` values and unwraps the `Some(t)` to just the value inside. Then the map I put after that is to convert the `&str` to a `String`

ruari-almost: Ok got it - cheers Daniel.

xd009642:
So I'm really sorry about this but when I went to test the features stuff I looked to the docs to figure out some examples to test it with and in this section https://doc.rust-lang.org/cargo/reference/config.html#target noticed the following quote:

>  cfg values come from those built-in to the compiler (run rustc --print=cfg to view), values set by build scripts, and extra --cfg flags passed to rustc (such as those defined in RUSTFLAGS). Do not try to match on debug_assertions or Cargo features like feature="foo".

I've also been looking at cargo metadata and can't see a way to get the cfg's from the build script. So I think if you just revert the feature handling stuff, I can begin to review this PR as the final solution. Also if a project is added to the tests/data directory that has a .cargo/config you can add an integration test to make sure it works. For that you can either do a project that uses something like simd that needs a special flag (`-C target-cpu=native` for simd) or just put something like `--cfg=foo` in the rustflags, and use the public function to get the rust flags and check the resulting string contains `--cfg=foo`. Both are acceptable although I'd personally prefer the first for simplicity :smile: 

Again sorry for getting you to add the unnecessary feature, if it's any consolation from a brief once over the code looked all good :smile:

ruari-almost:
> So I'm really sorry about this but when I went to test the features stuff I looked to the docs to figure out some examples to test it with and in this section https://doc.rust-lang.org/cargo/reference/config.html#target noticed the following quote:
> 
> > cfg values come from those built-in to the compiler (run rustc --print=cfg to view), values set by build scripts, and extra --cfg flags passed to rustc (such as those defined in RUSTFLAGS). Do not try to match on debug_assertions or Cargo features like feature="foo".
> 
> I've also been looking at cargo metadata and can't see a way to get the cfg's from the build script. So I think if you just revert the feature handling stuff, I can begin to review this PR as the final solution. Also if a project is added to the tests/data directory that has a .cargo/config you can add an integration test to make sure it works. For that you can either do a project that uses something like simd that needs a special flag (`-C target-cpu=native` for simd) or just put something like `--cfg=foo` in the rustflags, and use the public function to get the rust flags and check the resulting string contains `--cfg=foo`. Both are acceptable although I'd personally prefer the first for simplicity 😄
> 
> Again sorry for getting you to add the unnecessary feature, if it's any consolation from a brief once over the code looked all good 😄

No problem Daniel.

src/cargo.rs
```diff
@@ -432,6 +434,88 @@ pub fn rustdoc_flags(config: &Config) -> String {
...
+
+        Some(rustflags_in_file.join(" "))
+    } else {
+        None
+    }
+}
+
+fn look_for_rustflags_in(path: &Path) -> Option<String> {
+    let mut config_path = PathBuf::from(&path);
+    config_path.push("config");
```
xd009642, line 479: Just realised you can simplify this a bit by doing `let mut config_path = path.join("config");`

xd009642:
Just came back to give this a once over to refresh myself on it and it's all looking good. The minor simplification I mentioned previously and a test are all it needs currently :smile:

src/cargo.rs
```diff
@@ -554,10 +637,13 @@ pub fn rust_flags(config: &Config) -> String {
     }
     handle_llvm_flags(&mut value, config);
     if let Ok(vtemp) = env::var(RUSTFLAGS) {
+        value.push_str(vtemp.as_ref());
         lazy_static! {
             static ref DEBUG_INFO: Regex = Regex::new(r#"\-C\s*debuginfo=\d"#).unwrap();
         }
         value.push_str(&DEBUG_INFO.replace_all(&vtemp, " "));
+    } else {
+        value.push_str(gather_config_rust_flags(config).as_ref());
```
xd009642, line 646: Okay so move the lazy_static! block outside of the if statement, and then you can do 

```rust
let vtemp = gather_config_rust_flags(config);
value.push_str(&DEBUG_INFO.replace_all(&vtemp, " "));
```

src/cargo.rs
```diff
@@ -554,10 +637,13 @@ pub fn rust_flags(config: &Config) -> String {
     }
     handle_llvm_flags(&mut value, config);
     if let Ok(vtemp) = env::var(RUSTFLAGS) {
+        value.push_str(vtemp.as_ref());
```
xd009642, line 640: This line should be removed as we push the vtemp value altered by the regex only

ruari-almost: So the then block is now empty?  Am unsure what the resultant code should look like? Do we still care about RUSTFLAGS being set in the environment?

xd009642: The block still has `value.push_str(&DEBUG_INFO.replace_all(&vtemp, " "));` in it. Remember, if RUSTFLAGS is set then the users config.toml isn't read, so we don't want to gather the rustflags in that instance

It will be:

```rust
    lazy_static! {
        static ref DEBUG_INFO: Regex = Regex::new(r#"\-C\s*debuginfo=\d"#).unwrap();
    }
    if let Ok(vtemp) = env::var(RUSTFLAGS) {
       value.push_str(&DEBUG_INFO.replace_all(&vtemp, " "));
   } else {
        let vtemp = gather_config_rust_flags(config);
        value.push_str(&DEBUG_INFO.replace_all(&vtemp, " "));
     }
```

ruari-almost: Ah OK -we need the new DEBUG_INFO malarkey in both the then and the else.  Thanks.

tests/mod.rs
```diff
@@ -272,6 +273,27 @@ fn cargo_home_filtering() {
...
+    let mut config = Config::default();
+
+    let restore_dir = env::current_dir().unwrap();
+    let test_dir = get_test_path("rustflags");
+    env::set_current_dir(&test_dir).unwrap();
+    config.manifest = test_dir;
+    config.manifest.push("Cargo.toml");
+
+    let (_, ret) = launch_tarpaulin(&config, &None).unwrap();
+    env::set_current_dir(restore_dir).unwrap();
```
xd009642, line 288: ah sorry I've lead you astray! This line should be deleted as you restore the directory later

PR #579, opened 2020-10-02
xmac94x/extend logging

xMAC94x:
Add tracing for the error i discovered where tarpaulin hangs.
Will make debugging easier for the minimal example i am about to push


xd009642:
So I don't know if it's just me, but this doesn't print any additional logs in your minimal example?

xMAC94x:
for me it prints (when used with `-v` or `--debug`). But i think we can close this PR and instead include your `cargo_hang` branch instead, as it fixes the hanging situation better
ah you already merged, perfect.

Issue #566, opened 2020-09-21
how to check coverage for only files in the workspace?


KarthikNedunchezhiyan:
Hi, i am running tarpaulin command `cargo tarpaulin --all` inorder to check the coverage results.

tarpaulin.toml
```
[all]
verbose = true
debug = false
workspace = true
exclude-files = ["src/sparkdev/**/*"]
out = ["Html"]
```

I expected to take coverage for only files inside the workspace, but tarpaulin also includes the files outside of the workspace like below

```
|| ../../../usr/include/x86_64-linux-musl/ctype.h: 0/3
|| ../../../usr/include/x86_64-linux-musl/endian.h: 0/9
|| ../../../usr/include/x86_64-linux-musl/math.h: 0/4
```

Any idea on this?

xd009642:
I probably only apply the workspace filtering rules on rust source files, l'll work to fix the filtering. In the meantime you could use some variant of `--exclude-files` maybe `--exclude-files */usr/*` (I think those wildcards should work)

xd009642:
Oh I forgot to ask, do you have a public project I can  try or do you know which dependencies are bringing in those headers? Just so I can test my fix :smile:

KarthikNedunchezhiyan:
Hi @xd009642 , Great to hear this. I have very huge dependencies list so, i am really not sure what makes it to include this. Is there any way to find this, kindly let me know i will try that 😃 .

Additional context: It is reproducing in 0.14.3 but not in 0.12.4

xd009642:
Hmm you could always run cargo tree and paste the output here, I should be able to figure it out from thatb(though might take some time if there's truly a lot)

KarthikNedunchezhiyan:
ya sure, below is the output of `cargo tree -p yyyyy`

```
yyyyy v0.1.0 (/workspace/xxxxx/src/yyyyy)
├── channel v0.1.0 (/workspace/xxxxx/src/channel)
│   └── nix v0.18.0
│       ├── bitflags v1.2.1
│       ├── cfg-if v0.1.10
│       └── libc v0.2.76
│   [dev-dependencies]
│   └── mio v0.7.0
│       ├── libc v0.2.76 (*)
│       └── log v0.4.11
│           └── cfg-if v0.1.10 (*)
├── clap v2.33.3
│   ├── ansi_term v0.11.0
│   ├── atty v0.2.14
│   │   └── libc v0.2.76 (*)
│   ├── bitflags v1.2.1 (*)
[...]

PR #592, opened 2020-10-14
Rearrange is_coverable_file ordering in test_loader

xd009642:
* Rearrange the logic to be correct all the time with header includes
* is_coverable_file now check if file is in project directory as well not just if it's in bad directories
* Add tests


xd009642:
Fixes #566

PR #604, opened 2020-10-26
Initial version of running build binaries via tarp

xd009642:
~~Currently a bit clunky need to do `cargo tarpaulin --command Build --bin <bin_name>`. I should be able to fix the `--bin` defaults so it's not needed.~~ Just need to make sure I don't change anything that breaks it for tests.

So you should be able to (for a project with a single binary) run it with multiple sets or args and see how much you cover like so:

```
[run_help]
command = "build"
release = true
args = ["--help"]

[run_cmd_1]
command = "build"
release = true
args = ["tarpaulin", "--fail-under", "50", "--root", "test_project"]
```

Further testing still needed


xd009642:
I'm hoping that cargo will release it wouldn't have to rebuild the project for any but the first run, then for longer CLI apps they can run through some scenarios get coverage and then analyse logs or something for the pass/fail of the tests

xd009642:
This also coincided well with #610 so just going to merge now

PR #614, opened 2020-11-05
Add avoid-cfg-tarpaulin flag

krisukox:
I created this pull request in order to avoid "cannot find attribute" error that comes from `#[cfg_attr(tarpaulin, skip)]` in some old dependencies.
In the project I use `glutin` that uses `wayland-commons v0.21.13`. When I want to check coverage using tarpaulin I get:

> "Error: cannot find attribute \`skip\` in this scope"

`--avoid-cfg-tarpaulin` flag removes `--cfg=tarpaulin` from the `RUSTFLAGS`.
Do you agree to add such a flag?
Thanks in advance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ file.
 
 ## [Unreleased]
 ### Added
-- Added `--avoid-cfg-tarpaulin` flag to remove --cfg=tarpaulin from the RUSTFLAGS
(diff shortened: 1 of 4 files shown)


xd009642:
Yeah, I've endeavoured to get projects that haven't upgraded to upgrade via PRs (when I've spotted them) but some haven't yet or don't plan on releasing an update for a while. If you just update the changelog I'll approve and merge this.

krisukox:
Done

xd009642:
Awesome, thanks 👍

PR #630, opened 2020-11-25
Exit with with a non-zero code if cargo does so

### LINKED ISSUE #627: Tarpaulin ignores non-zero exit code from a custom build script
I have a project where I have a custom build script, and when the build script exits with a non-zero, tarpaulin exits with a 0. [This](https://github.com/Ch00k/orthanc-cli/pull/11/checks?check_run_id=1441380815) pipeline (the `run_tests` step) illustrates the problem. [`build.rs`](https://github.com/Ch00k/orthanc-cli/blob/1a0a0b4824b5ab58a06af14ead00b7b74b9feea8/build.rs) fails because it cannot write into a directory, that does not exist.

I hope this much information is enough, but please let me know if I need to provide any more, or to create a minimal example that would illustrate the issue.

Ch00k:
Fixes cargo exit code handling. Fixes #627


xd009642:
Ah thank you, I'm away right now so have limited time to work on tarpaulin. I'll review this later tonight :+1:

xd009642:
Also if you update the changelog as well please :smile:

xd009642:
Cool all looks good to me :+1:

Issue #549, opened 2020-08-29
LLVM coverage instrumentation


xd009642:
So when llvm coverage instrumentation is stable on rust we can look at removing the ptrace based coverage completely and instead relying on llvm and rustc to help. This has a few advantages

1. None of the weirdness of ptrace 
2. More accurate results
3. First class support in the compiler - less chance of behaviour changing in nightly
4. Linker issues, large target folders and slowness should be greatly mitigated
5. Easier route to get branch/condition and other coverage metrics by coupling llvm_lines and source analysis

In this instance tarpaulin should make the whole experience still nicer because:

1. Handles calling cargo and right flags
2. Doc tests coverage without needing to remember the nightly flags
3. results collation/report generation
4. tagging unused generics that won't be instrumented

OSX TODOs:

- [x] doc_coverage::doc_test_coverage
- [x] doc_coverage::doc_test_env
- [x] doc_coverage::doc_test_panics
- [x] doc_coverage::doc_test_panics_workspace
- [x] failures::error_if_test_fails
- [x] follow_exes_down
- [x] handle_forks
- [x] issue_966_follow_exec
- [x] kill_used_in_test
- [x] match_expr_coverage
- [x] rustflags_handling
- [x] test_types::only_test_coverage

Windows TODOs

- [x] failures::error_if_test_fails
- [x] follow_exes_down
- [x] issue_966_follow_exec
- [x] match_expr_coverage
- [x] rustflags_handling
- [x] picking_up_shared_objects
- [x] doc_coverage::doc_test_panics
- [x] doc_coverage::doc_test_panics_workspace

Linux TODOs

- [x] doc_coverage::doc_test_panics
- [x] failures::error_if_test_fails
- [x] handle_forks
- [x] follow_exes_down
- [x] issue_966_follow_exec
- [x] match_expr_coverage
- [x] rustflags_handling
- [x] test_types::only_test_coverage

General TODO

- [x] Coverage behaviour with loops

xd009642:
https://hammer-vlsi.readthedocs.io/en/stable/CoverageMappingFormat.html some docs and the rust issue I keep having to find again  https://github.com/rust-lang/rust/issues/34701

caemor:
Is there already a branch where we can try this?

xd009642:
Not yet, given my current work schedule I'm hoping to have something early January

xd009642:
[...]

PR #653, opened 2020-12-29
llvm-cov support pt1

xd009642:
Work in progress branch for #549. Ideally I'd like to have the parsing of the files done in rust so we don't need to get users to pull in all the llvm tools just to parse a report. Some things that need figuring out:

1. Running multiple tests what happens to the default.profraw?
2. Tests that exec other binaries in the project, what happens to the default.profraw?
3. Parsing profraw and converting to the internal TraceMap
4. Detecting if the nightly version is new enough for llvm-cov - probably need to check the date in the version string :grimacing:


xd009642:
can confirm doc tests work as expected, two reports generated as there isn't a no-run option for doctests

PR #687, opened 2021-02-16
Fixes #601

### LINKED ISSUE #601: "0.00% coverage" reported when rustflags contains `-C debuginfo=0`
In https://github.com/rust-lang/crates.io/pull/2951 I've added https://github.com/actions-rs/tarpaulin to the CI pipeline of crates.io, but unfortunately it currently reports "0.00% coverage" and I'm out of ideas of what I might be doing wrong 😞 

The project definitely does have tests so the coverage should be at least above zero. Any help would be greatly appreciated! 🙏

xd009642:
This detects if the user has set -Cdebuginfo in their rustflags and
removes this as tarpaulin sets it and requires it to get debug info for
traces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ file.
 - Added `--color` option matching cargo arg
 - `--follow-exec` option making exec tracing non-default
 - `--jobs` option matching the one in cargo test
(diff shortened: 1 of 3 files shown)


MoreTacos:
This is a comment

PR #760, opened 2021-05-16
Update README.md

pinkforest:
* Adds a disclaimer about Cargo environment variables not available as documented in Cargo

https://github.com/rust-lang/cargo/issues/8842
https://github.com/alexliesenfeld/httpmock/issues/45

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,9 +171,7 @@ In order to achieve this it sets the following environment variables when execut
 - **CARGO_PKG_VERSION**   - _From Cargo.toml_
 - **LLVM_PROFILE_FILE**   - _Used for LLVM coverage_
 
-### Cargo Limitations
-
-Since tarpaulin is a custom Cargo sub-command, the regular Cargo environment has been constructed by tarpaulin.
+### Cargo Manifest
 
 In order for tarpaulin to construct the Cargo environment correctly, tarpaulin needs to find Cargo.toml by either:
 - Using *--root* or *--manifest-path* or
--- a/README.md
+++ b/README.md
@@ -172,15 +172,14 @@ Tarpaulin sets the below environment vars under test:
 
 **Cargo Limitations**
 
-Since tarpaulin is a custom Cargo sub-command, the regular Cargo environment has been constructed by tarpaulin via 
-guesswork which may in rare cases result in unexpected behaviour under test.
+Since tarpaulin is a custom Cargo sub-command, the regular Cargo environment has been constructed by tarpaulin.
 
 In order for tarpaulin to construct the Cargo environment correctly, tarpaulin needs to find Cargo.toml by either:
 - Using *--root* or *--manifest-path* or
 - By invoking cargo from the current working directory holding the Cargo.toml manifest or
 - By invoking cargo from a child directory under the parent holding the Cargo.toml
 
-If Cargo does not find any Cargo.toml from using either of above methods the run will error and exit.
+If Cargo does not find any Cargo.toml from using either of above methods the run will error "cargo metadata" and exit.
 
 Several RFCs are open in rust-lang to expose [more of these](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-3rd-party-subcommands) directly in order to avoid the issues arising out of this.
 
--- a/README.md
+++ b/README.md
@@ -161,10 +161,10 @@ This ensures that tarpaulin will be built with the same rust version as the rest
 
 ### Environment Vars
 
-Tarpaulin in best effort basis sets the below environment vars under test:
+Tarpaulin sets the below environment vars under test:
 
 - **RUST_BACKTRACE**      - _When --verbose flag is used_
-- **CARGO_MANIFEST_DIR**  - _Cargo.toml Path guessed from the current working or parent directory_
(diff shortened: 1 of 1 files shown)


xd009642:
`CARGO_MANIFEST_DIR` is set by tarpaulin when the tests are run so they behave the same as if you run directly with `cargo test`. We also make sure to set as many useful vars as possible: https://github.com/xd009642/tarpaulin/blob/53dfa4223ef01a09178b32bdf326dc9cb7ecdb1c/src/process_handling/mod.rs#L106-L121 

Admittedly, some may be missing or if cargo behaviour changes then tarpaulin will behave differently, so while I'm not opposed to the warning I don't think the current phrasing is quite accurate. Are there any variables you've noticed aren't set or are set incorrectly by tarpaulin that prompted this PR?

pinkforest:
My bad I should have expanded description on those reference items and done more looking around the code as I was assuming it is the sub-command default where it's not exposed downwards.

Thanks for clarifying

It seems to be that Doctests are not evaluating CARGO_MANIFEST_PATH correctly while as Tests works.

I've opened issue: https://github.com/xd009642/tarpaulin/issues/761

Re-producible: https://github.com/pinkforest/test_tarpaulin_env

**Changing this PR**

Instead of warning unexposed environment variables this PR should probably document the exposed env's per code

And need to figure out why Doctests does not expose CARGO_MANIFEST_PATH and PR that

Now also need to figure out how to test the sub-command run within tarpaulin repo test instead of now evaluating under installed sub-command like the reproduction of at test_tarpaulin_env which has both Tests and Doctests run types.

Tarpaulin tests/workspaces.rs already tests under workspace for Tests but probably run a Doctest there as well

xd009642:
So I had a bit of time to look at tarpaulin again and did #764 feel free to have a look, and yeah if the documentation can change to reflect the desire of tarpaulin to match the env your tests run under so they run equivalently :+1:

pinkforest:
@xd009642 I committed a section about Environment variables and some blurb about it's limitations under this PR for README.md should be merge'able now? Thx

pinkforest:
Woops I made an error assuming it's just best effort guesswork when it was just a bug where #764 blew past checker.

I clarified on last commit about that the manifest must be found.

README.md
```diff
@@ -159,6 +159,30 @@ installation and massively reducing the install time on CI.
 When using the [Nix](https://nixos.org/nix) package manager, the `nixpkgs.cargo-tarpaulin` package can be used.
 This ensures that tarpaulin will be built with the same rust version as the rest of your packages.
 
+### Environment Vars
+
+Tarpaulin sets the below environment vars under test:
```
xd009642, line 164: change "vars" to "variables"

README.md
```diff
@@ -159,6 +159,30 @@ installation and massively reducing the install time on CI.
...
+
+- **RUST_BACKTRACE**      - _When --verbose flag is used_
+- **CARGO_MANIFEST_DIR**  - _Path to Cargo.toml From --root | --manifest-path or guessed from the current or parent directory_
+- **CARGO_PKG_NAME**      - _From Cargo.toml_
+- **CARGO_PKG_AUTHORS**   - _From Cargo.toml_
+- **CARGO_PKG_VERSION**   - _From Cargo.toml_
+- **LLVM_PROFILE_FILE**   - _Used for LLVM coverage_
+
+**Cargo Limitations**
```
xd009642, line 173: If you use Heading 4 here so `#### Cargo Limitations`

README.md
```diff
@@ -159,6 +159,30 @@ installation and massively reducing the install time on CI.
...
+
+Since tarpaulin is a custom Cargo sub-command, the regular Cargo environment has been constructed by tarpaulin.
+
+In order for tarpaulin to construct the Cargo environment correctly, tarpaulin needs to find Cargo.toml by either:
+- Using *--root* or *--manifest-path* or
+- By invoking cargo from the current working directory holding the Cargo.toml manifest or
+- By invoking cargo from a child directory under the parent holding the Cargo.toml
```
xd009642, line 180: I think something like "By invoking cargo from a sub-directory within the project" would be nicer phrasing. Also for these points if you capitalise cargo to make it consistent with the other usages

README.md
xd009642: Additionally, I think it's worthwhile to add a bit more run in, just to mention that tarpaulin tries to set up an equivalent environment to when you run your tests via cargo test i.e. 

> When tarpaulin runs your tests it strives to run them in the same environment as if they were ran via `cargo test`. In order to achieve this it sets the following environment variables when executing the test binaries:

This is only a rough wording I haven't given it too much thought. If you think you have a nicer phrasing I'm open to that as well :smile:

xd009642 (changes requested):
Just a few tweaks and nits, but overall looking good :+1:

README.md
pinkforest: Thanks incorporated

pinkforest: I tend to use the env everywhere for some reason, good catch :)

README.md
pinkforest: Thanks I thought about ### earlier but you are right :)

pinkforest:
Thanks incorporated all the feedback, good catches :)

xd009642:
Nice LGTM :+1:  (I'll skip CI as it's just readme changes)

PR #764, opened 2021-05-18
Propagate CARGO_MANIFEST_DIR to doctest

xd009642:
@pinkforest I had a look at the code and figured out why it wasn't getting the env. On my local test I recreated and showed it fixed and added a test but feel free to try it out and see if there's anything else that should be propagated that's missing


pinkforest:
Works in my world too now :) LGTM :1st_place_medal: 

Thanks!

PR #767, opened 2021-05-27
Fix cleaning for wrong directory

AtkinsChang:
I think it there is no Easter egg for `cargo` to put release build in `${CARGO_TARGET_DIR}/debug`


xd009642:
Ah thanks for that :sweat_smile:

mathstuf:
Hmm. After this change, all of my builds are now done from scratch (see #777). Any idea why that might be?

mathstuf:
Seems to have been a red herring; this change just exposed that problem for me.

PR #779, opened 2021-06-16
gitignore: ignore target as a symlink

mathstuf:
Reapplication of 6aaf36519b0864e01a25311101394c4ada65cb18 which seems to have been undone in 5a9e96dd8002de953502292341ab45fb4ba154dc.


mathstuf:
Thanks!

PR #802, opened 2021-08-11
feat(config): log config parsing

CHr15F0x:
This change allows to see config parse errors. Entire conf vector is also dumped for additional verification if `trace` enabled.

The motivation for this PR is that it is quite difficult to get the configuration file correctly quickly as parse errors will not be reported. I encountered this problem while trying to write a config file, where I needed to combine several `cargo test` runs into a single coverage report.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -20,6 +20,8 @@ use tracing::{error, info, warn};
 mod parse;
 pub mod types;
 
+#[derive(Debug)]
+
 pub struct ConfigWrapper(pub Vec<Config>);
 
 /// Specifies the current configuration tarpaulin is using.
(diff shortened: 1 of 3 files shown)


src/config/mod.rs
```diff
@@ -20,6 +20,8 @@ use tracing::{error, info, warn};
 mod parse;
 pub mod types;
 
+#[derive(Debug)]
+
```
xd009642, line 24: NIT Can you remove this extra line break? Also if you update the changelog with this change, then it looks good to go :+1:

CHr15F0x: Sorry about that one :facepalm:, I'll fix it right away.

xd009642:
Cool thanks :+1:

PR #818, opened 2021-09-02
Use core from existing mask when setting processor affinity

### LINKED ISSUE #817: Segmentation fault when running inside a container
**Describe the bug**
I recently encountered a segmentation fault error when running tarpaulin (version 0.18.0) with multi-threaded tests inside an LXC container. This looked very similar to issue #190, as it appeared randomly. I realized that tarpaulin shows the following warning every time when starting a test:
```
WARN cargo_tarpaulin::process_handling::linux: Failed to set processor affinity EINVAL: Invalid argument
```
I tried to prevent any actual parallel execution by setting the processor affinity manually with `taskset -ac 0 cargo tarpaulin`, which also failed with the message `Invalid argument`.
Looking at the affinity mask of existing processes, it showed that all processes in this container were running with the mask `0xefb5420050` (containing 16 bits set to one for 16 cores). This might be caused by the fact that only some cores are forwarded inside the container.

Setting the mask to one or two of those cores made the segfault disappear (e.g. with `taskset -ac 4,6 cargo tarpaulin`). Setting it to three or more cores though made the problem reappear. 

Looking at the tarpaulin code at https://github.com/xd009642/tarpaulin/blob/develop/src/process_handling/linux.rs#L85 the warning and the issues might be fixable by checking an existing affinity mask and setting it to only one of those existing cores. I'll try this out today or tomorrow to see if it consistently solves the segfault issues.

Although I was wondering why this affinity setting is needed in the first place. It seems to me that tarpaulin only works correctly if no code is actually executed in parallel, enforced by the processor affinity. Is this an intended limitation or just a workaround for deeper issues?

**To Reproduce**
I used the following example to produce the segfaults. They happen at around 25% of the time. [lib.rs](https://github.com/xd009642/tarpaulin/files/7092164/lib.rs.txt)

kaenganxt:
As discussed in #817, this change sets the processor affinity to the lowest core id of the existing affinity mask. This will solve cases where the core id 0, which was previously used, is not available (e.g. in virtualized environments).
(fixes #817)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ file.
 - Now handles string values for rustflags in .cargo/config not just a list of values
 - INTERNAL If llvm coverage is enabled and test binary can't be loaded start with empty `TraceMap`
 - Config parse errors are logged
+- Setting the processor affinity now uses an existing core from the initial affinity mask instead of defaulting to the first one (see issue #817)
 
 ### Removed
 
--- a/src/process_handling/linux.rs
+++ b/src/process_handling/linux.rs
@@ -83,9 +83,19 @@ fn disable_aslr() -> nix::Result<i32> {
 }
 
 pub fn limit_affinity() -> nix::Result<()> {
-    let mut cpu_set = CpuSet::new();
-    cpu_set.set(0)?;
     let this = Pid::this();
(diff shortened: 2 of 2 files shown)


xd009642:
Just tested and recreated issue and proved this fixes it :+1:

PR #853, opened 2021-10-11
Unify the code for ImplItemMethod and ItemFn

teromene:
This PR unifies the code used to check coverage between `impl` methods and regular functions.
It aims at removing inconsistent behaviour such as this one:
```rust
struct Test {
    value1: usize,
    value2: usize,
    value3: usize,
}

impl Test {
    fn test_value() -> Option<Test> {
        println!("Hello world !");
        Some(Test {
            value1: 7,
            value2: 9,
            value3: 13,
        })
    }
}

fn test_value() -> Option<Test> {
    println!("Hello world !");
    Some(Test {
        value1: 7,
        value2: 9,
        value3: 13,
    })
}

#[test]
fn test() {
   let tv = Test::test_value().unwrap();
   assert_eq!(tv.value1, 7);

   let tv = test_value().unwrap();
   assert_eq!(tv.value1, 7);
}
```
leading to this coverage report:
image
With this PR, the extra fields in the method `Test::test_value` are properly ignored, as they are in the function.

This PR breaks the `source_analysis::tests::cover_generic_impl_methods` test: indeed, elements like
```rust
struct GenericStruct<T>(T);
impl<T> GenericStruct<T> {
    fn hw(&self) {
        println!("hello world");
    }
}
```
are no longer marked as coverable.
This behaviour however is coherent with existing behaviour of generics, and highlights an underlining bug:
The following code (note the empty line at the start) reports correctly that the code is not covered:
```rust

fn hw(data: String) {
        println!("hello world");
} 
```
image

but a generic version marks the code as not coverable (again, an empty line is required or the program behaves as expected) :
```rust

fn hw<T>(data: T) {     
        println!("hello world");
} 
```
image

### DIFF
--- a/src/source_analysis/items.rs
+++ b/src/source_analysis/items.rs
@@ -16,7 +16,7 @@ impl SourceAnalysis {
                     analysis.ignore_tokens(i);
                 }
                 Item::Mod(ref i) => self.visit_mod(&i, ctx),
-                Item::Fn(ref i) => self.visit_fn(&i, ctx),
+                Item::Fn(ref i) => self.visit_fn(&i, ctx, false),
                 Item::Struct(ref i) => {
                     let analysis = self.get_line_analysis(ctx.file.to_path_buf());
                     analysis.ignore_tokens(i);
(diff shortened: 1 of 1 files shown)


teromene:
I understand, but the behaviour is terribly inconsistent: Taking my example with the generic function `hw<T>(data: T)`, a single empty line before the declaration marks the function as not coverable.

xd009642:
Yeah we don't have windows support yet, it's mainly just there to prove it builds and I haven't added in anything platform specific without a cfg for when future windows support starts working.

I'll have a review of this after work :+1:

PR #870, opened 2021-11-04
Fixes #869

### LINKED ISSUE #869: Generating coverage xml of a library that has examples produces line-rate=NaN
**Describe the bug**
Running with arguments `--out Xml` on a library containing an examples folder produces a section in the cobertura.xml:
```
        <package name="examples" line-rate="NaN" branch-rate="0" complexity="0">
            <classes></classes>
        </package>
```
This causes the overall line-rate to be NaN too.

**Expected behavior**
Coverage for the examples should just be 0 if there are no tests.

xd009642:
Add guard for if there are no lines in a package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ file.
 - Support for `#[no_coverage]` to exclude expressions from coverage results
 
 ### Changed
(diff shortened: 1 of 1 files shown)


xd009642:
I'm a idiot who didn't commit my change...

PR #874, opened 2021-11-05
Fix the build on aarch64-unknown-linux-gnu

JohnTitor:
`Persona` and `INT` are cfg-gated thus the build on aarch64-unknown-linux-gnu fails. I think the functionality should also work on AArch64 so added `aarch64` to the `cfg`s.

### DIFF
--- a/src/process_handling/breakpoint.rs
+++ b/src/process_handling/breakpoint.rs
@@ -4,7 +4,7 @@ use nix::unistd::Pid;
 use nix::{Error, Result};
 use std::collections::HashMap;
 
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
 const INT: u64 = 0xCC;
 
 /// Breakpoint construct used to monitor program execution. As tarpaulin is an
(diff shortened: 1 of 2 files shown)


src/process_handling/breakpoint.rs
```diff
@@ -4,7 +4,7 @@ use nix::unistd::Pid;
 use nix::{Error, Result};
 use std::collections::HashMap;
 
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
```
xd009642, line 7: I was googling around to see what the equivalent instruction was in ARM to see if it's actually the same instruction and so far I've just got this on the LLDB mailing list: https://www.mail-archive.com/lldb-dev@lists.llvm.org/msg09058.html but it doesn't seem that they use the same opcode for software breakpoints...

xd009642:
Also have you tried this on aarch64? One requirement for adding a new architecture would be having it added to the CI so if you also do that as part of this PR

src/process_handling/breakpoint.rs
xd009642: Yeah you'll want to find the bytes for the `BKPT` instruction. Also, if it's longer than a byte the `Breakpoint` type needs to change to handle varying length instructions depending on arch

JohnTitor:
Ohhh, the problem is harder than I thought, I'm going to investigate further and close this for now. Thanks for pointing out!

xd009642:
No worries, I created an issue for this https://github.com/xd009642/tarpaulin/issues/876

Issue #210, opened 2019-02-06
Ignore vendored crates


xd009642:
Found when working on #190 with tmplgen, tarpaulin looks to all the vendored crates stored for offline building for coverage data as well. Take into account local repositories and ignore vendored dependencies by default - potentially add a flag to include them for those using for internal closed source projects.

xd009642:
So I've been revisiting some issues I hadn't looked at in a while and realised this one is essentially sorted by `--exclude-files`. Also cargo-vendor uses vendor as the default folder name but you can change it and some user may have a project in the workspace called vendor.

So in the name of keeping things simple and explicit I'm just going to close this issue. Though if anyone has an disagreements feel free to comment and I'll think them over

PR #883, opened 2021-11-28
Exclude files already when walking source tree

epontan:
It is known that `tarpaulin` will look for coverage data in vendored
crates if `cargo-vendor` is used (#210). It is reasonably simple to
exclude the vendored crates though the `--exclude-files` option.
However, all of the sources is still collected even if they are not
treated as covered. This is quite obvious when using the `Html` report
format for example which becomes huge because it contains the crate
sources as well as all the vendored crates' source files.

To avoid this, apply the excluded files filter at the earliest point,
i.e., when walking the crate source tree and finding all the rust source
files.

---

Small example:
```bash
cargo new tarpaulin-vendor-reprod
```
`Cargo.toml`:
```toml
  [package]
  name = "tarpaulin-vendor-reprod"
  version = "0.1.0"
  edition = "2021"

  [dependencies]
  heapless = "0.7"
```
Vendoring:
```bash
cargo vendor > .cargo/config
```

Sample code:
```rust
use heapless::Vec;

fn foo<const N: usize>(bar: Vec<u32, N>) -> u32 {
    bar.iter().sum()
}

#[test]
fn test_foo() {
    let v: Vec<u32, 8> = Vec::from_slice(&[1, 2, 3]).unwrap();
    assert_eq!(foo(v), 6);
}
```

Run tarpaulin:
```bash
cargo tarpaulin --out Html --exclude-files vendor
```

Html report becomes 9MB (with two runs for comparison) compared to 140KB with the patch.

### DIFF
--- a/src/path_utils.rs
+++ b/src/path_utils.rs
@@ -64,7 +64,7 @@ pub fn is_coverable_file_path(
     ignorable_paths && is_part_of_project(e, root.as_ref())
 }
 
-pub fn get_source_walker(config: &Config) -> impl Iterator<Item = DirEntry> {
+pub fn get_source_walker(config: &Config) -> impl Iterator<Item = DirEntry> + '_ {
     let root = config.root();
     let target = config.target_dir();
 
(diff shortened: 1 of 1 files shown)


xd009642:
Thanks :+1:

PR #931, opened 2022-01-05
remove a couple of allocations

danieleades:
this PR removes quite a few `collect` calls at the expense (probably) of readability. In particular it complicates a few function signatures my introducing opaque types.

I'm not sure how much of a performance bottleneck these are anyway, but i leave it up to you what you want to do with this.


xd009642:
All of the changes so far look pretty sensible so I'm happy merging once there's something that works :+1:

xd009642:
Problem is given the most of the time will be spent in compilation or waiting on ptrace signal/stop/delivery loop it's very hard to definitively measure performance :sweat_smile:

xd009642:
I thought that might be the case, but just in case it wasn't :upside_down_face:

Issue #954, opened 2022-02-09
Try rusty-fork for integration tests


xd009642:
https://crates.io/crates/rusty-fork rusty-fork may allow us to run the tests without `--test-threads 1` by having the ptrace listens being in a fork child. This could potentially speed up CI and make it easier for new tarpaulin devs to run the tests.

PR #956, opened 2022-02-12
Apply some tweaks based on cargo-mutants

xd009642:
Firmed up the branch analysis tests and fixed one issue (for when it
hopefully moves forward in the future). Fixed a gap in source_analysis
tests. Minor tweak to CARGO_HOME test to make it a bit cleaner

Fixes https://github.com/xd009642/tarpaulin/issues/954


xd009642:
Notes on tests that could be added:

1. Run doc tests project, modify to change line of doc test and the doc test comment and then run again. Ensure results aren't confused.

Things I can't think of a way to add:

1. Some of the "untested" features involve tweaking the args passed to the test binary. I don't have a way of introspecting hose args but I could maybe break out into another function and unit test them :thinking:

PR #958, opened 2022-02-14
Fix cargo install hint in README

flxo:
Fix trivial typo in readme that bit me when copy and casting the install command.


flxo:
Sorry - copied the wrong field for `binstall`

PR #993, opened 2022-04-20
Refactoring

zhiburt:
Hi there

I was (once again) reading some code here and there then noticed `clippy` issues + some duplication.

So I decided to do something

Take care.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,34 +72,23 @@ pub fn setup_logging(color: Color, debug: bool, verbose: bool) {
 }
 
 pub fn trace(configs: &[Config]) -> Result<TraceMap, RunError> {
+    let logger = create_logger(configs);
     let mut tracemap = TraceMap::new();
+    let mut ret = 0;
     let mut tarpaulin_result = Ok(());
-    let mut ret = 0i32;
-    let logger = if configs.iter().any(|c| c.dump_traces) {
-        Some(EventLog::new(configs.iter().map(|x| x.root()).collect()))
-    } else {
-        None
-    };
-    let mut bad_threshold = None;
+    let mut bad_threshold = Ok(());
     for config in configs.iter() {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -115,8 +104,11 @@ pub fn trace(configs: &[Config]) -> Result<TraceMap, RunError> {
             }
         }
     }
+
     tracemap.dedup();
-    if let Some(bad_limit) = bad_threshold {
+
+    // It's OK that bad_threshold, tarpaulin_result may be overwritten in a loop
```
zhiburt, line 110: Is this correct behavior?
I did not changed it juts left a comment that it is.

xd009642:
So are you happy if I merge this and then do the fail threshold in another PR? :eyes: Just wanted to make sure you're not working on it before I hit merge

zhiburt:
Initially I did not planned to do it.
And I was typing a corresponding response here.

But then I tried.....

And didn't managed that :disappointed: 

So you can take over :)

_______

A note:

Could you check this test https://github.com/zhiburt/tarpaulin/blob/1722d63a99a0017bc07babab75b338f471c4c9bb/tests/failures.rs#L88-L104?

I run 2 configs guessing that 2 of them must be run.
But I see only output of 1 config.

I am not very versed in the code base and did not investigate things deeply.

xd009642:
So I see the output of both configs, and if I remove the failing config the test fails, if I remove the good config it passes. I'll merge this though, issue has been raised in https://github.com/xd009642/tarpaulin/issues/995

xd009642:
test do_nothing is in success, and test it_works are in failing. Both project tests are in there. And the type is an error type instead of reporting it and ending with success - that looks as expected

xd009642:
ah see you spotted that in an edit :sweat_smile:

PR #1101, opened 2022-10-01
Add 2-decimal point display precision for the report coverage % delta

zedseven:
This PR just adds 2-decimal point precision to the display for the coverage % delta and one error `Display` impl that also displays percentages.


xd009642:
Looks good  to me :+1:

PR #1102, opened 2022-10-02
Add `rust-cache` to the CI workflow.

zedseven:
When submitting #1101, I noticed the CI took a long time.

On all of my projects, I use https://github.com/marketplace/actions/rust-cache, which caches the compiled dependencies so that only the final executable is rebuilt on every push/PR. This should significantly speed up successive CI runs. (less so for nightly builds because the cache is invalidated daily)

I noticed on the Linux job, there was a `cargo clean` step that wasn't done on Windows or Mac. Because that would defeat the point of caching compiled dependencies, I removed it.

I also noticed the `RUST_BACKTRACE` environment variable was set individually, and moved that to a global workflow-level environment variable, so it applies everywhere. If that was a mistake, and it was applied individually for a reason, I can move it back.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,6 +6,9 @@ on:
       - "*"
   pull_request:
 
+env:
+  RUST_BACKTRACE: 1
+
 jobs:
   linux:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


xd009642:
Cool thanks for that, I admit I haven't taken a lot of time speeding up CI - it used to take 40 minutes on travis so getting it to what it was on github actions was more than enough for me initially :sweat_smile:

Issue #1129, opened 2022-11-07
Release has unconventional name


nstinus:
The latest release is named  cargo-tarpaulin-0.22.0-travis.tar.gz  which does not show the target triple.
It seems to be a x86_64-unknown-linux-gnu arch. It should be named cargo-tarpaulin-0.22.0-x86_64-unknown-linux-gnu.tar.gz

This has the bad consequence of tricking `cargo binstall` into thinking that it is fine for a musl executable.

xd009642:
This has now changed, whole release stuff is reworked

Issue #1130, opened 2022-11-07
Offer a release for x86_64-unknown-linux-musl target


nstinus:
For users living on old distros (e.g. CentOs 7), a musl release would be very helpful!

xd009642:
So I've tried a release today and it looks like it didn't work if you don't mind having a look as the person who did the PR :eyes:  https://github.com/xd009642/tarpaulin/actions/runs/3498060423

nstinus:
https://github.com/xd009642/tarpaulin/pull/1136

xd009642:
Okay that didn't fix it but I've gone through 5 iterations of CI "fixes" and it looks like it will work next time there's a release. Github actions seems to only checkout current commit so it failed on the command to get the current tag but built fine. 

I'll try and smash through a few issues over the weekend to justify another release to test it :smiling_face_with_tear:

nstinus:
I'm sorry it was so difficult, to be honest, I have zero experience with actions... Thank you!

xd009642:
It's okay, I should have been smarter than trusting sensible looking changes in CI would "just work". And as long as it adds something useful for users it's worth the hassle in the end :sweat_smile:

PR #1131, opened 2022-11-07
Fix x86_64-unknown-linux-gnu release filename and add one for x86_64-unknown-linux-musl

nstinus:
(Hopefully) fixes issues #1129 and #1130.


xd009642:
Looks like it should work to me, I'll merge and we'll find out next release (CI failure doesn't matter just some issue I need to solve on mac :cry: )

PR #1150, opened 2022-11-29
Initial work on fixing unc paths and reports

xd009642:


xd009642:
This is caused by https://github.com/rust-lang/rust/issues/42869

PR #1174, opened 2023-01-09
Update release workflow

taiki-e:
As said in https://github.com/taiki-e/install-action/pull/42#issuecomment-1372184541, I would like to help improve release staff. (If work on this is in progress on your end and you do not need this patch, feel free to close this PR.)

This patch updates the release workflow and makes the release work on x86_64 linux-gnu/linux-musl/macos/windows and aarch64 linux-gnu/linux-musl/macos. The following are specific changes:

- Add x86_64/aarch64 linux-musl
  - Closes https://github.com/xd009642/tarpaulin/issues/1130
- Remove aarch64 windows.
  - ~~aarch64 linux fails to build [^1]: https://github.com/taiki-e/tarpaulin/actions/runs/3875026523/jobs/6606999714~~ EDIT: fixed in https://github.com/xd009642/tarpaulin/pull/1174/commits/5476833acd2cf994ebecba6949d29aeb74a857e7
  - aarch64 windows doesn't support profiler_builtins
- Enable vendored-openssl feature
  - At least on linux-musl we cannot build without it. [^1]
  - For now, I enabled this for all targets, but it is possible to control whether this is enabled or disabled per target.
- Use the latest taiki-e/create-gh-release-action and actions/checkout
- Update binstall metadata to match new archive names.

I tested this with a fork by temporarily disabling the configs around `cargo publish`; here is generated release: https://github.com/taiki-e/tarpaulin/releases/tag/0.23.1
I also confirmed that the binaries created in the above test pass install-action's CI in all environments: https://github.com/taiki-e/install-action/actions/runs/3875547860/jobs/6608244922

cc https://github.com/taiki-e/install-action/pull/42
fyi @thomcc @NobodyXu 

[^1]: [These targets are included in the test matrix in rust.yml](https://github.com/xd009642/tarpaulin/blob/5f5e566e173d34afe7919b480b40336c4aaf6537/.github/workflows/rust.yml#L21-L22), but the job is actually testing the host target instead of these targets because [the --target option is not passed to cargo](https://github.com/xd009642/tarpaulin/blob/5f5e566e173d34afe7919b480b40336c4aaf6537/.github/workflows/rust.yml#L33-L39).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,8 @@ jobs:
             os: ubuntu-latest
           - target: aarch64-apple-darwin
             os: macos-latest
+          - target: universal-apple-darwin
+            os: macos-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/upload-rust-binary-action@v1
         with:
           bin: cargo-tarpaulin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: taiki-e/create-gh-release-action@v1.3.0
+      - uses: taiki-e/create-gh-release-action@v1
         with:
           changelog: CHANGELOG.md
         env:
@@ -20,7 +20,7 @@ jobs:
   crates:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: publish package to crates
         run: |
           cargo package
@@ -31,9 +31,12 @@ jobs:
     needs:
       - create-release
     strategy:
+      fail-fast: false
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
           - target: x86_64-apple-darwin
             os: macos-latest
@@ -41,16 +44,17 @@ jobs:
             os: windows-latest
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-latest
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
           - target: aarch64-apple-darwin
             os: macos-latest
-          - target: aarch64-pc-windows-msvc
-            os: windows-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
       - uses: taiki-e/upload-rust-binary-action@v1
         with:
-          bin: rust-sandbox
+          bin: cargo-tarpaulin
           target: ${{ matrix.target }}
+          features: vendored-openssl
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ rusty-fork = "0.3.0"
 
 [package.metadata.binstall]
 pkg-url = "{ repo }/releases/download/{ version }/cargo-tarpaulin-{ target }.tar.gz"
-bin-dir = "cargo-tarpaulin"
+bin-dir = "cargo-tarpaulin{ binary-ext }"
 pkg-fmt = "tgz"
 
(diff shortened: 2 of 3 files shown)


taiki-e:
> But maybe that's because they're building on target not via cross and my detecting of x64_64 in my build.rs doesn't work when cross compiling 🤔

Ah, #[cfg] in the build script refers to the host target on which the build script is run, so the following line seems to be causing the problem when cross-compiling.

https://github.com/xd009642/tarpaulin/blob/5f5e566e173d34afe7919b480b40336c4aaf6537/build.rs#L13-L14

xd009642:
Yeah that what I was concerned about... Is there a way in build scripts to access the _actual_ target architecture you're compiling for :thinking:

taiki-e:
> Is there a way in build scripts to access the _actual_ target architecture you're compiling for 🤔

Cargo sets those information via [`CARGO_CFG_*` environment variables](https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts).

taiki-e:
Fixed build scripts and re-added aarch64 linux/macos to the workflow. 

I can confirm that the cross-compilation issue on aarch64 linux has been fixed: https://github.com/taiki-e/tarpaulin/actions/runs/3876026314/jobs/6609363516

.github/workflows/release.yml
```diff
@@ -31,26 +31,30 @@ jobs:
...
       - create-release
     strategy:
+      fail-fast: false
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
           - target: x86_64-apple-darwin
```
NobodyXu, line 41: Perhaps it's a good idea to support universal-apple-darwin (I know there's no official name for this, but binstall does use that name) by using lipo

taiki-e: Did you mean to just add universal-apple-darwin to the matrix? Or did you mean to replace aarch64-apple-darwin and x86_64-apple-darwin with universal-apple-darwin?

I added universal binary support to upload-rust-binary-action, so either way is ok.

NobodyXu: > I added universal binary support to upload-rust-binary-action, so either way is ok.

Yes, this was what I am thinking ahout.

xd009642:
If you just give me a ping when this is ready for final review/merge as I notice there's still an unresolved discussion RE mac. Looking forward to giving this a spin :tada:

.github/workflows/release.yml
taiki-e: @NobodyXu 
> > I added universal binary support to upload-rust-binary-action, so either way is ok.
> 
> Yes, this was what I am thinking ahout.

Um, I guess you quoted the wrong sentence?

xd009642: I would guess @NobodyXu meant universal-apple-darwin just from context :thinking:

NobodyXu: Oh yes, I didn't receive nofitication for @taiki-e 's reply.

taiki-e: When I saw NobodyXu's review, I wasn't sure if NobodyXu meant "distributing aarch64 and x86_64 while also distributing universal" or "distributing universal instead of aarch64 and x86_64", so my question was to know which NobodyXu meant.

NobodyXu: I meant distributing aarch64 and x86_64 while also distributing universal binaries.

Having all available is useful since aarch64 and x86_64 are smaller than universal binary, but users might want universal binary so that it works on any mac target.

.github/workflows/release.yml
```diff
@@ -31,26 +31,30 @@ jobs:
...
             os: macos-latest
           - target: x86_64-pc-windows-msvc
             os: windows-latest
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-latest
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
           - target: aarch64-apple-darwin
             os: macos-latest
```
taiki-e, line 50:
```suggestion
            os: macos-latest
          - target: universal-apple-darwin
            os: macos-latest
```

.github/workflows/release.yml
taiki-e: Thanks for the clarification. I added universal-apple-darwin to the matrix.

taiki-e:
This should now be ready for review.

.github/workflows/release.yml
```diff
@@ -31,26 +31,32 @@ jobs:
...
-          - target: aarch64-pc-windows-msvc
-            os: windows-latest
+          - target: universal-apple-darwin
+            os: macos-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
```
taiki-e, line 55:
```suggestion
      - uses: actions/checkout@v3
      - uses: dtolnay/rust-toolchain@stable
```

xd009642:
LGTM, thanks a lot for this :tada:

xd009642 (approved):

PR #1208, opened 2023-01-30
Clippy: was re-implementing find() call.

martinfrances107:
Clippy has a rules which warns when you are re-implementing a core function

in this case "find" was being implemented.

Here is a description of the rule [find rule](https://rust-lang.github.io/rust-clippy/master/index.html#manual_find)


xd009642:
LGTM - all the other clippy ones do as well so when CI passes I'll just quietly merge :smile:

PR #1225, opened 2023-02-24
More fixes for supplementary objects

### LINKED ISSUE #1156: No coverage for cdylib crate called from C/C++ executables
**Describe the bug**

I have a project (https://github.com/Luthaf/rascaline/, coverage branch, PR with CI is https://github.com/Luthaf/rascaline/pull/126) which is built as both a rust crate (tarpaulin works great here, thanks a lot!) and exports a C API through a cdylib crate (`rascaline-c-api`). The C API is then used to create a C++ (and Python) interface for the code.

The C and C++ API are tested as a normal C/C++ project, using CMake to first build the rust code, and then build C and C++ test executables. I made sure to pass the right flags to rustc (setting `RUSTFLAGS="-Cdebuginfo=2 --cfg=tarpaulin -Cinstrument-coverage -Clink-dead-code"`) when building the rust shared library. 

The LLVM-based coverage instrumentation seems to work fine, generating `.profraw` files (I added a small patch in https://github.com/Luthaf/tarpaulin/commit/c67eb17884510e06fe7fca0ce61fc15805367537 since the profraw files are generated inside a different directory), but then no coverage is reported for the c-api crate at all. I added a handful of `dbg!` inside cargo-tarpaulin and llvm-profparser, and as far as I understand the code it seems that the profraw files contain at least some coverage data.

**To Reproduce**

```bash
git clone https://github.com/Luthaf/rascaline
cd rascaline 
git checkout coverage
cargo tarpaulin --all-features --workspace --engine=llvm --test=c-api-tests --skip-clean --verbose
```

Here is the output I get: [tarpaulin-rascaline.txt](https://github.com/xd009642/tarpaulin/files/10167205/tarpaulin-rascaline.txt)

I'm wondering if there could be an issue with file name parsing or something like this, since all the path shown are relative to the `rascaline-c-api` crate, and not the root of the repository/root of the cargo workspace.

If you have a couple of pointer/thinks I could look for, I'm happy to continue debugging this and send a PR once it is fixed.

**Expected behavior**

Using an instrumented shared library from C or C++ should be able to collect and report coverage on the C API implementation in rust.

Luthaf:
Note: this PR targets the `fix/canonicalise-object` branch, not `develop`.

I need one other change to fully fix #1156 for me: the supplementary objects are only created in the middle of the test suite, so trying to access it before results in "file not found". I think the simplest solution is to check if the file exists before trying to parse it, informing the user if the file is skipped so that typos can be corrected.

### DIFF
--- a/src/event_log.rs
+++ b/src/event_log.rs
@@ -4,8 +4,12 @@ use crate::config::Config;
 use crate::ptrace_control::*;
 #[cfg(ptrace_supported)]
 use crate::statemachine::ProcessInfo;
+#[cfg(ptrace_supported)]
 use crate::statemachine::TracerAction;
-use crate::traces::{Location, TraceMap};
+#[cfg(ptrace_supported)]
+use crate::traces::TraceMap;
+
+use crate::traces::Location;
 use chrono::offset::Local;
 #[cfg(ptrace_supported)]
 use nix::libc::*;
--- a/src/statemachine/instrumented.rs
+++ b/src/statemachine/instrumented.rs
@@ -114,8 +114,26 @@ impl<'a> StateData for LlvmInstrumentedData<'a> {
                         let code = exit.code().unwrap_or(1);
                         return Ok(Some(TestState::End(code)));
                     }
-                    // Panics due to a todo!();
-                    let mut binaries = parent.extra_binaries.clone();
+
+                    let mut binaries = parent
+                        .extra_binaries
+                        .iter()
+                        .filter(|path| {
+                            // extra binaries might not exist yet and be created
+                            // later by the test suite
+                            if path.exists() {
+                                true
+                            } else {
+                                info!(
+                                    "Skipping additional object '{}' since the file does not exist",
+                                    path.display()
+                                );
+                                false
+                            }
+                        })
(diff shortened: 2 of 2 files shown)


Luthaf:
I ran rustfmt & force-pushed, tests where failing before

src/event_log.rs
```diff
@@ -4,8 +4,12 @@ use crate::config::Config;
 use crate::ptrace_control::*;
 #[cfg(ptrace_supported)]
 use crate::statemachine::ProcessInfo;
+#[cfg(ptrace_supported)]
 use crate::statemachine::TracerAction;
-use crate::traces::{Location, TraceMap};
+#[cfg(ptrace_supported)]
```
xd009642, line 9: I don't think this change is required? The TraceMap is still used for llvm coverage, it might trigger an unused warning though

Luthaf: TraceMap is just below, with an unconditional import. And yes, this is my second commit, just fixing some unused warnings. I can take it out if you prefer.

xd009642: nah it's fine was just making sure

Issue #1229, opened 2023-03-06
`--no-fail-fast` doesn't seem to work as expected in cargo workspace


gen-xu:
**Describe the bug**

In cargo workspace, when running tarpaulin with `cargo tarpaulin --no-fail-fast` would still fail at the first time it encounters a failed test executable. The `--no-fail-fast` only applies when there is a test failure within the package, but tarpaulin would quit earlier if the test executable exited with error, and tarpaulin is unable proceed to next test binary.

After read the code a bit, I think the reason is that the `get_test_coverage` https://github.com/xd009642/tarpaulin/blob/develop/src/lib.rs#L248 propagated the `RunError::TestFailed` inside the loop of all test binaries.

**To Reproduce**
Create two  or more crates in a cargo workspace, with one of them has a test that fails.

**Expected behavior**
`--no-fail-fast` should finish executing all tests inside cargo workspace regardlessly.

PR #1230, opened 2023-03-06
Fix cargo worksapce test with --no-fail-fast

gen-xu:
Addressing https://github.com/xd009642/tarpaulin/issues/1229

Allow workspace test binaries loop to continue if `--no-fail-fast` is provided


gen-xu:
I guess one question I have here is that when we have `--no-fail-fast` on, do we still want tarpaulin to output test coverage or not, and should it have 0 exiting code or non-0 exiting code? IMO it makes sense to have non-0 exiting code but output coverage but I would like to hear more opinions.

xd009642:
>  IMO it makes sense to have non-0 exiting code but output coverage but I would like to hear more opinions. 

That makes the most sense to me as well, so no dissenting opinions here :+1:

xd009642:
Hi @gen-xu mild poke do you plan on finishing this off sometime soon? If you're too busy I should hopefully have some free bandwidth to push it over the line next week

gen-xu:
> Hi @gen-xu mild poke do you plan on finishing this off sometime soon? If you're too busy I should hopefully have some free bandwidth to push it over the line next week

I think the efforts needed was than I expected so I have been leaving it off for a while. Yeah, if you have somed bandwidth for this it would be great. Thanks!

Issue #1220, opened 2023-02-15
`--ignore-tests` doesn't ignore `#[tokio::test]` or `#[async_std::test]`


xd009642:
As described in title.

xd009642:
Fixed in #1233

PR #1233, opened 2023-03-09
Added filtering of attributes that end with 'test' like tokio::test

Quessou:
Addressing https://github.com/xd009642/tarpaulin/issues/1220

Added the ability to ignore functions that are prefixed with an attribute ending with `test`.
Added matching automated tests.


xd009642:
Looks good to me, also good job on more complicated attributes such as `#[tokio::test(worker_threads = 1)]` as well!

PR #1287, opened 2023-05-21
Fix cargo worksapce test with --no-fail-fast

xd009642:


xd009642:
TODO still have non-zero exit code but show coverage

xd009642:
Fixes #1229

PR #1301, opened 2023-06-11
Simplify cargo version parsing

amoskvin:
Cargo from distro packages does not contain a date, which results in the version not getting parsed.

e.g.:
```
$ cargo --version
cargo 1.66.1
```
The date is not being used anywhere, so I just removed it.


xd009642:
LGTM! Thanks :smile:

PR #1319, opened 2023-07-03
Fix include-tests

vigneshs-12:
### Reason for this PR

This commit 343e912bf33a3872ff9f49d48686831d7ca701da made `ignore-tests` the default flag and added `include-tests` flag to reapply the old behaviour. But, the [ignore_tests](https://github.com/xd009642/tarpaulin/commit/343e912bf33a3872ff9f49d48686831d7ca701da#diff-b0ba3e54b6f5ef1fa6aed8075e352a68fda66067fbafa8421042a2347801338aR392) function will always return `true` even if `include-tests` is set. This is because of the logical OR operator `||` used.

### Description of changes

- Removed `ignore_tests` from `Config` as both `ignore_tests` and `include_tests` are inverse to each other and can be deferred from a single boolean. 
- Removed `set_ignore_tests` and added `set_include_tests`.
- Removed `ignore_tests` and added `include_tests`.
- Made required changes to use the new  `set_include_tests` and `include_tests`.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -38,9 +38,6 @@ pub struct Config {
     /// Flag to also run tests with the ignored attribute
     #[serde(rename = "ignored")]
     pub run_ignored: bool,
-    /// Flag to ignore test functions in coverage statistics
-    #[serde(rename = "ignore-tests")]
-    ignore_tests: bool,
     /// Ignore panic macros in code.
     #[serde(rename = "ignore-panics")]
     pub ignore_panics: bool,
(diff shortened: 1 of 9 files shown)


xd009642:
Looks mostly good, test failure seems to be down to some of the windows based conditional compilation (I guess just testing a windows specific feature)

vigneshs-12:
Fixed

xd009642 (approved):

xd009642:
Thanks for this, LGTM!

PR #1374, opened 2023-09-09
added option to fail immediately after single test fails

### LINKED ISSUE #1373: Add option to fail immediately after single test fails
**Describe the feature**
Add option to fail immediately after single test fails, currently when running cargo tarpaulin if test fails during run only at the end it gives error, but all other tests are still runned
```
Error: "Test failed during run"
```

Add option to fail immediately after single test fails, this would make running faster, because all tests would not needed to be ran

0xMimir:
When running code coverage on bigger projects, if test fails, tests for all other libraries and binaries are still runned, it would be nice to have a option to fail immediately after single test fails, like `cargo test` does.

Closes #1373

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -201,6 +201,9 @@ pub struct ConfigArgs {
     /// CI server being used, if unspecified tarpaulin may automatically infer for coveralls uploads
     #[arg(long, value_name = "SERVICE")]
     pub ciserver: Option<Ci>,
+    /// Option to fail immediately after a single test fails
+    #[arg(long)]
+    pub fail_immediately: bool,
     /// Arguments to be passed to the test executables can be used to filter or skip certain tests
     #[arg(last = true)]
     pub args: Vec<String>,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ From 2019 onwards, all notable changes to tarpaulin will be documented in this
 file.
 
 ## [Unreleased]
+### Added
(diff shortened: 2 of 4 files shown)


src/args.rs
```diff
@@ -201,6 +201,9 @@ pub struct ConfigArgs {
     /// CI server being used, if unspecified tarpaulin may automatically infer for coveralls uploads
     #[arg(long, value_name = "SERVICE")]
     pub ciserver: Option<Ci>,
+    /// Option to fail immediately after single test fails
```
xd009642, line 204: Grammar nit: "fail immediately after a single test fails"

src/main.rs
```diff
@@ -24,7 +24,8 @@ fn print_env(seen_rustflags: HashMap<String, Vec<String>>, prefix: &str, default
 
 fn main() -> Result<(), String> {
     let args = CargoTarpaulinCli::from_args();
-
+    // println!("{:#?}", args.config);
```
xd009642, line 27: Remove the commented out lines

xd009642 (changes requested):
So two small tidiness comments, and if you can update the changelog something like as follows:

```
## [Unreleased]
### Added
- Added `--fail-immediately` flag to abort execution the moment the first test failure occurs

### Changed
```

0xMimir:
Fixed typos, removed commended lines and added changes to `CHANGELOG.md`

xd009642:
Ah and cargo fmt is failing in CI so if you can run `cargo fmt` and push the resulting changes as well

0xMimir:
I have added changes from cargo fmt

PR #1395, opened 2023-09-30
Fix default ValueEnum casing

### LINKED ISSUE #1392: `--out` casing changed with release 0.27
This may not be a bug, but it's not mentioned the release notes or in the commit message that introduced this so I'm not sure it's intentional.

The casing of the arguments of the `--out` parameter changed with the migration to Clap v4 in https://github.com/xd009642/tarpaulin/commit/d1bd48ea2ac1356ea3087b5d2d197ee58702dbd0. This means that old scripts using `--out Html` will fail (expecting `--out html`).
This is changed because [`ValueEnum` defaults to `kebab-case`](https://docs.rs/clap/latest/clap/_derive/index.html#valueenum-attributes).

This is not necessarily a bug, but it is an undocumented breaking change.

If this is not intentional (or even if it is), I think it would be great to support both casing for a while to ease the transition, maybe with a deprecation warning. This makes it easier for scripts to migrate to the new version progressively.

MiniaczQ:
Fixes #1392

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -181,16 +181,16 @@ pub struct ConfigArgs {
     #[arg(short = 'Z', value_name = "FEATURES", num_args = 0..)]
     pub unstable_features: Vec<String>,
     /// Output format of coverage report
-    #[arg(long, short, value_enum, value_name = "FMT", num_args = 0..)]
+    #[arg(long, short, value_enum, value_name = "FMT", num_args = 0.., ignore_case = true)]
     pub out: Vec<OutputFile>,
     /// Coverage tracing backend to use
-    #[arg(long, value_enum, value_name = "ENGINE")]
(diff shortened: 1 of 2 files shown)


xd009642:
surprised it's not rename_all `pascalCase` like in serde :thinking: . Also I think the old behaviour might have accepted lower case and pascal case? Could you just check that for me (or if I do before I'll edit comment/review to mention it)

MiniaczQ:
I'll check after work

MiniaczQ:
It's `PascalCase` in serde, old version accepts all cases, I'll check how that can be done

MiniaczQ:
I patched it so it allows any casing

xd009642:
Amazing! I'll cut a new release after work then :tada:

Issue #1411, opened 2023-10-23
Return an error if the coverage is decreasing


nikgul:
**Describe the feature**
It would be nice if tarpaulin would return an error if the coverage percentage is decreasing, kinda like you set a threshold with --fail-under with a set percentage.

xd009642:
Closed as wontfix for reason mentioned in PR (replicated here)

> I've thought about this more and I've decided it's not something I'm looking to merge into tarpaulin, there's some complexity in the behaviour RE times when fail-decreasing doesn't fail if coverage decrease and also the vast majority of users are running in CI and then this type of functionality is provided by other services like coveralls etc.

PR #1412, opened 2023-10-23
Fail decreasing

nikgul:
This pull request implements the `--fail-decreasing` command line option, which allows tarpaulin to return an error if the coverage percentage has decreased, similar to `--fail-under` which fails the run if the coverage is under a certain threshold.

Current feature set implemented:
- If --fail-under is used by it self, it should behave like before
- If --fail-decreasing is used by it self, it will return an error if the coverage percentage is decreasing.
- If --fail-under and --fail-decreasing is used together is works as follows, if the coverage is below the threshold, it error out if the coverage is decreasing, if the coverage is above the threshold, the coverage can increase and decrease freely.

I have had a quick look at the tests, but are uncertain how and if I should make a test case for this, becaues it would require multi step tests, a first run, one increasing, one returning the same coverage, and one where it is decreasing.

This fixes: #1411

Note: I'm new to rust, so if you have any changes you want made, there could be some terminoligy, that i'm not fluent in or familier with yet :-)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@ use crate::errors::*;
 use crate::event_log::*;
 use crate::path_utils::*;
 use crate::process_handling::*;
-use crate::report::report_coverage;
+use crate::report::{report_coverage, report_delta};
 use crate::source_analysis::{LineAnalysis, SourceAnalysis};
 use crate::test_loader::*;
 use crate::traces::*;
@@ -177,6 +177,18 @@ fn check_fail_threshold(traces: &TraceMap, config: &Config) -> Result<(), RunErr
     }
 }
 
+fn check_fail_decreasing(traces: &TraceMap, config: &Config) -> Result<(), RunError> {
+    if config.fail_decreasing {
+        let delta = report_delta(config, traces);
+        if delta < 0.0f64 {
+            let error = RunError::CoverageDecreasing(delta);
+            error!("{}", error);
+            return Err(error);
+        }
+    }
+    Ok(())
+}
+
 pub fn run(configs: &[Config]) -> Result<(), RunError> {
     if configs.iter().any(|x| x.engine() == TraceEngine::Llvm) {
         let profraw_dir = configs[0].profraw_dir();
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -229,9 +251,48 @@ pub fn report_tracemap(configs: &[Config], tracemap: TraceMap) -> Result<(), Run
     Ok(())
 }
 
-fn report_coverage_with_check(c: &Config, tracemap: &TraceMap) -> Result<(), RunError> {
-    report_coverage(c, tracemap)?;
-    check_fail_threshold(tracemap, c)
+fn check_coverage(delta: f64, config: &Config, trace: &TraceMap) -> Result<(), RunError> {
+    let threshold = check_fail_threshold(trace, config);
```
xd009642, line 255: You can make this function a lot simpler by doing just:

```rust
	check_fail_threshold(trace, config)?;
        if config.fail_under.is_none() {
	    check_fail_decreasing(delta, config)?;
        }
         Ok(())
```

I think this covers all the feature space, you'd just have to move the error log into `check_fail_decreasing`

xd009642 (changes requested):
So one comment to simplify a function down, otherwise it looks fine to me. 

Another thing you can do is add an integration test. Just take one of the integration test projects where coverage can change (i.e. one with features and you can enable more features). And then set up the combinations of fail-under and fail-decreasing. I have a bunch of tests for fail-under in `tests/failure_thresholds.rs` where these tests would fit in.

src/lib.rs
nikgul: I'm not sure that would be correct with the current feature set ?
If I understand the change request correctly, then check_coverage would return the threshold error, regardles if the coverage is increasing, could you read the commit message on the last commit with the fail-under integration and see if those cases are still coveret, which I don't think they are.

Have forgotten to update the pull request descrition

xd009642: Well if you set `--fail-under` then I would expect it to always fail if below that coverage amount even if the coverage is increasing because you haven't met the threshold.

nikgul: My thought was, that when you start to use tarpaulin, that you would be able to use it at once, and you set your end target from the start and don't have to correct it later, by providing kind of a "soft start", so that you don't have to provide all test cases at once, but you will have to provide tests for all new code.
Once you reach the threshold, then --fail-decreasing is basically not used anymore with the code implemented, if used in combination with --fail-under, where it takes over.
The user is always welcome to use them individually if the combined functionality is not what they want.

xd009642:
Just at work so will look properly later, but I use `set_profraw_dir` on the target to stop tests clobbering each other.

nikgul:
I tried that, but couldn't get it to work, ended up using config.set_target_dir() instead, which works nicely.

src/lib.rs
xd009642: Hmm I think that just adds confusion to the flag. If you don't want to see the behaviour at the start of a project you can just omit it. Though come to think of it most users use tarpaulin in CI and this flag won't do anything then anyway as they typically won't cache the previous tarpaulin results.

xd009642:
I've thought about this more and I've decided it's not something I'm looking to merge into tarpaulin, there's some complexity in the behaviour RE times when fail-decreasing doesn't fail if coverage decrease and also the vast majority of users are running in CI and then this type of functionality is provided by other services like coveralls etc.

Issue #1433, opened 2023-12-12
Consider adding line numbers and making the file path float in the HTML report


HadrienG2:
Upon noticing something in the HTML coverage report, one may want to jump to the matching location of the corresponding source file in their code editor in order to adjust the code or to automatically generate a test skeleton using an IDE feature.

I think this switch would be easier with two HTML output UI tweaks:

- Make the header that shows the file path a floating box that stays visible no matter how far you've scrolled down the page.
- Add a line number gutter on the left so it's clear which line of the file we are looking at.

xd009642:
Those do sound like good improvements, I'm fairly busy preparing for a conference talk i'm doing next year, plus day job. But I have next week off and will try to take a look into doing this with my limited FE skills :+1:

sbatial:
@xd009642 out of curiosity and because I have some frontend experience so I'd be glad to work on the line numbers as well:
- do you specifically want to have the report as a single html file?
- any specific reason (apart from the former) for using react (and not something static/rust native for example)?

xd009642:
I've not really done much frontend before, the HTML reports were actually a PR I accepted and I've tried to rope other people into solving issues with because of my dislike of trying to make javascript do what I want. 

The only real requirement I have is that it be runnable without a CDN to fetch JS files - hence why the `react-dom-production.min.js` is included in the repo. Because that's a no go for some industry users. I could probably be persuaded to move away from single file - even though I think it keeps everything nice and contained for my own usage

sbatial:
Right, ok.

I agree, I was/am also quite happy with the single file approach.

I see what I can do.

xd009642:
@HadrienG2 there's now line numbers and the floating header thanks to @sbatial if you want to check it out. CI is just churning through the release process so there should be something on crates.io later on today

HadrienG2:
Thanks a lot, looks perfect to me!

PR #1437, opened 2023-12-25
Make header stick to the top even when scrolling down

sbatial:
This should fix the first part of #1433.

It keeps the top bar stuck to the top of the window when scrolling down.

The crucial part `sticky` is pretty well supported across browsers. 
https://caniuse.com/css-sticky


xd009642 (approved):
LGTM, thanks for this (and the wait over holiday period)

PR #1467, opened 2024-01-31
Add missing --target to make use of ${{ matrix.target }}

kpcyrd:
I noticed the build is currently failing on 32bit x86 when trying to [package cargo-tarpaulin for Alpine Linux](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/59413).

In the github actions workflow I noticed there's a test matrix defined for i686, but without `--target` it only runs for the default target (x86_64).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,9 +32,9 @@ jobs:
       - name: cache
         uses: Swatinem/rust-cache@v2
       - name: test
-        run: cargo test
+        run: cargo test --target ${{ matrix.target }}
       - name: ignored test
-        run: cargo test -- --ignored || true
+        run: cargo test --target ${{ matrix.target }} -- --ignored || true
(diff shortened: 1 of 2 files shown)


xd009642:
So it looks like there's something odd with github actions and the targets, do you want any help looking into this?

kpcyrd:
Feel free to pick it up, the repo is currently configured to require manual CI runs until I have at least 1 commit on the codebase, it's setup like this by default to protect people using self-hosted runners:

image

I'd need to create a PR on my own repository first, make sure the pipelines pass, then create a PR upstream.

At the moment it seems to be missing `rustup target add ${{ matrix.target }}`.

xd009642:
well CI has stopped running just as I think I may have got it working so I'm just going to merge this into a tarpaulin feature branch and work from there. Thanks for spotting this I'll update on the issue when I make some progress

Issue #1477, opened 2024-02-17
Add --package as alias for --packages


tommy-gilligan:
- `cargo test` has `--package` and `-p`
- `cargo tarpaulin` has `--packages` and `-p`

I think it would be nice if `cargo tarpaulin` also had `--package` as an alias to make it easier going between these tools.

PR #1478, opened 2024-02-17
Add --package alias to mimic `cargo test` interface

tommy-gilligan:
- cargo test has `--package` and `-p`
- cargo tarpaulin has `--packages` and `-p`

I think it would be nice if cargo tarpaulin also had `--package` as an alias to make it easier going between these tools.

https://github.com/xd009642/tarpaulin/issues/1477

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -124,7 +124,7 @@ pub struct ConfigArgs {
     #[arg(long)]
     pub workspace: bool,
     /// Package id specifications for which package should be build. See cargo help pkgid for more info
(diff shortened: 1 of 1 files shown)


xd009642 (approved):
Thanks for this, I'll merge it once CI passes :+1:

xd009642:
Ah yeah changing the help message is probably more of a feature to request of clap (if it's that big an issue ofc). 

Merging now without the CI pass because it seems the failure is a nightly thing anyway (that I should now go and look at)

Issue #351, opened 2020-02-23
Misses in coverage


xd009642:
Mention any false positives or false negatives in code coverage in this issue with an example where it occurred and it will be added to the list!

- [ ] An `else` only line https://codecov.io/gh/Smithay/wayland-rs/src/master/wayland-commons/src/socket.rs#L33 
- [ ] Single logical lines split across multiple ones
- [ ] match statement with enum or `*self`  https://github.com/xd009642/tarpaulin/issues/258
- [ ] https://github.com/xd009642/tarpaulin/issues/270 
- [ ] macros in function bodies https://codecov.io/gh/simd-lite/simdjson-rs/src/master/src/stage2.rs#L133 
- [ ] constant or static values can be uncovered https://codecov.io/gh/simd-lite/simdjson-rs/src/master/src/stage2.rs#L320 
- [ ] unsafe blocks marked as uncovered https://codecov.io/gh/simd-lite/simdjson-rs/src/master/src/lib.rs#L285 
- [ ] `const` initialisations aren't covered.
- [ ] general chained method calls (sometimes can differ in free functions versus methods see https://github.com/xd009642/tarpaulin/issues/238 )
- [ ] `PhantomData`
- [ ] Inconsistent coverage with macros like `assert_eq!`
- [ ] Not sure how well inline `asm!` works

jamestwebber:
It looks like my coverage report is counting the lines _in_ tests as part of coverage (including one miss, somehow). See e.g. [here](https://codecov.io/gh/czbiohub/bcl2fastr/pull/73/src/src/common/write_fastq.rs#L517) if you scroll down to the lines 500+. Not sure whether that's intentional or not but seems a bit odd (and the miss is a false negative I guess)

xd009642:
There's some weirdness with macros, I need to figure out the best way to handle them. Also if you want to exclude test code you can use `--exclude-tests` and all lines in anything with the `#[test]` attribute or `#[cfg(test)]` will be removed from the results

pwoolcoc:
Just curious, why is `--ignore-tests` not the default? Are people really writing tests for their tests? I can't imagine a situation where I'd want my test code to be counted as part of the coverable source code

xd009642:
Honestly just because as a feature it came later and no ones asked for the default to be switched around, I've got no opposition to switching the default behaviour. I also definitely don't think people are writing tests for their tests (or at least I hope not...) 

[...]

PR #1503, opened 2024-04-04
Fix `include-test` config to include non-std test

tyrone-wu:
Motivation:
The original issue is sorta described here: https://github.com/xd009642/tarpaulin/issues/351#issuecomment-2036231706. 
But tl;dr version is when `include-tests` is set, `#[tokio::test]` is not covered in the coverage report. 

Changes in this PR:
Added an additional condition to `visit_fn` to check whether the `PathSegment`s ends in `test`, such as `#[tokio::test]` or `#[some_fancy_crate::test]`. 

I tested this tarpaulin version against the issue that I was originally encountering, and it seems to fix it.  
Before where `#[tokio::test]` was ignored: https://app.codecov.io/gh/kube-rs/kube/blob/cov-tweak/kube-runtime%2Fsrc%2Fcontroller%2Ffuture_hash_map.rs#L85
After where it's now included: https://app.codecov.io/gh/tyrone-wu/kube/blob/main/kube-runtime%2Fsrc%2Fcontroller%2Ffuture_hash_map.rs#L85

### DIFF
--- a/src/source_analysis/items.rs
+++ b/src/source_analysis/items.rs
@@ -103,7 +103,8 @@ impl SourceAnalysis {
         for attr in &func.attrs {
             if let Ok(x) = attr.parse_meta() {
                 let id = x.path();
-                if id.is_ident("test") {
+                if id.is_ident("test") || id.segments.last().is_some_and(|seg| seg.ident == "test")
+                {
                     test_func = true;
                 } else if id.is_ident("derive") {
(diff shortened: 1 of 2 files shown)


xd009642:
Cool thanks for this, don't worry about the beta and nightly failures, it's something I'm currently (slowly) working on :sweat:

PR #1583, opened 2024-07-19
Fixes #1582

### LINKED ISSUE #1582: Make coveralls api integration an optional feature
Today there is no way to disable the coveralls integration when compiling tarpaulin.

This features is the only reason that `openssl` needs to be brought in. Openssl is one of the main pain points when attempting to compile cargo tarpaulin. 

For some reason even with the `vendored-openssl` feature enabled, tarpaulin still relies on a dynamically linked openssl1.1 when building in a docker container. Being able to disable the coverall API integration which we do not need would make this problem go away.

xd009642:
This PR aims to make the coveralls api crate and git2 optional to avoid needing to pull in openssl unnecessarily.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,6 +32,8 @@ jobs:
           components: rustfmt
       - name: cache
         uses: Swatinem/rust-cache@v2
+      - name: check_non_default
+        run: cargo check --no-default-features
       - name: test
         run: cargo test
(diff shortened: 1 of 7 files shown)


sosthene-nitrokey:
This looks like it's working for me

xd009642:
Cool, in that case I'll merge it some point after work (might be tomorrow morning) and probably cut a release when I do

PR #1592, opened 2024-07-29
Support `#[coverage(off)]` to exclude results

### LINKED ISSUE #1591: Support the new `coverage(off)` attribute format
The [`coverage_attribute`](https://doc.rust-lang.org/unstable-book/language-features/coverage-attribute.html) has switched from `no_coverage` to `coverage(off)` in https://github.com/rust-lang/rust/pull/114656. As far as I can tell, tarpaulin only supports the former.

zecakeh:
The [coverage_attribute](https://doc.rust-lang.org/unstable-book/language-features/coverage-attribute.html) feature has switched from `no_coverage` to `coverage(off)` in https://github.com/rust-lang/rust/pull/114656.

This adds support for the new format.

I wasn't sure if the old format should be removed right away. It might need a deprecation first.

Closes #1591.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 From 2019 onwards, all notable changes to tarpaulin will be documented in this
 file.
 
+## [Unreleased]
+### Added
+- Support for `#[coverage(off)]` to exclude expressions from coverage results
+
 ## [0.31.0] 2024-07-22 
 ### Added
 - Ability to remove coveralls from the build making openssl optional.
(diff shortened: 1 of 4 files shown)


xd009642:
I try to go rather conservative on deprecations, so if the compiler still supports `#[no_coverage]` then we'll keep it in. If it doesn't keep it in if it's low effort otherwise assess and decide

zecakeh:
`no_coverage` is not supported anymore by the compiler.

xd009642:
Cool we'll keep it in for now I'll just add a note about it after merge thanks for this looks good to me  :+1:

xd009642 (approved):

PR #1606, opened 2024-08-20
Remove HashMap println! in set_functions

kpcyrd:
This looks like a left-over debug `println!` but it's causing issues in [`cargo2junit`](https://github.com/johnterickson/cargo2junit) when using it together with `cargo-tarpaulin`:

```sh
RUSTC_BOOTSTRAP=1 cargo tarpaulin --out Xml --skip-clean --target-dir target-tarpaulin \
    -- -Z unstable-options --format json --report-time | cargo2junit > report.xml
```

`cargo-tarpaulin` is writing it's own report file, but the rustc output of `--format json` is also processed by `cargo2junit` at the same time (to avoid running `cargo test` twice).

`cargo2junit` is trying to detect rustc output by looking for lines starting with `{`, this is also the case for `std::fmt::Debug` of the HashMap (while not valid json), causing the `cargo2junit` parser the fail.

This patch solves this by removing the debug `println!`.

### DIFF
--- a/src/traces.rs
+++ b/src/traces.rs
@@ -169,7 +169,6 @@ impl TraceMap {
     }
 
     pub fn set_functions(&mut self, functions: HashMap<PathBuf, Vec<Function>>) {
-        println!("{:?}", functions);
         self.functions = functions;
     }
 


xd009642:
Sorry about that, I'll get a new release out this evening with this patch in!

kpcyrd:
No problem, thanks for responding/merging so quickly! :)

I'm subscribed to releases on this repo, I'm going to submit the new version to Alpine as soon as it's up.

kpcyrd:
0.31.2 is now available in Alpine edge: https://pkgs.alpinelinux.org/package/edge/community/x86_64/cargo-tarpaulin

PR #1633, opened 2024-10-10
feat: try to get path to config-file from `CARGO_TARPAULIN_CONFIG_FILE`

sivizius:
With this commit, tarpaulin will try to get the path to the configuration file (usually `tarpaulin.toml`) from the environment variable `CARGO_TARPAULIN_CONFIG_FILE` first, before proceeding with trying `tarpaulin.toml` and `.tarpaulin.toml`. The motivation of this PR is, that I have multiple repositories with the same tarpaulin-config. With https://direnv.net/, I am able to set environment variables, which would allow me to set `CARGO_TARPAULIN_CONFIG_FILE`, thus avoiding duplicate configurations. Symlinks are not really an option with repositories.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 From 2019 onwards, all notable changes to tarpaulin will be documented in this
 file.
 
+## [0.31.3] 2024-10-10
+### Added
+- The `CARGO_TARPAULIN_CONFIG_FILE` environment variable may be used to set the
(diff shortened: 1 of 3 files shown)


xd009642:
Had a brief look and it seems fine, I've been on holiday for the week after eurorust so I'll properly look at this tomorrow and merge it if it's all good 👍

xd009642 (approved):

PR #1652, opened 2024-11-26
Try to fix linker paths issue with deps using native deps

### LINKED ISSUE #1297: target/debug/deps folder not propagated in LD_LIBRARY_PATH
Using tarpauling 0.25.1.
I have a crate building a shared library (.so) with c - code (build.rs).
I have another rust-only crate exposing FFI code to that library (calls made in both directions, rust calls into C code, rust expose C FFI functions to be called from the library).
The rust crate has in it's Cargo.toml set a dependency to the dynamic C library:

**[dev-dependencies]
my-library.workspace = true**

In the rust crate I have a test in the tests folder, loading the C library dynamically using libloading (e.g. dlopen).
Running this with cargo test works fine, with tarpaulin the library is not found.
I inspected the LD_LIBRARY_PATH in both cases, when running tarpaulin the:

**<repofolder>/target/debug/deps**

folder was missing (but not when running cargo test) and is the reason for the failure.
I am not sure if I have missed anything that I should have given to tarpaulin or if it's a problem not specific to tarpaulin but the inconsistency here bring me the think it's an issue.

**To Reproduce**
Build a c library using build.rs, it's enough with one C function in there.
Build a rust crate using the function in the C library.
Create a test using libloading that loads up the C library, calling get to resolve the C library's function.
Call it.

**Expected behavior**
The C library should have been found by the test code when running tarpaulin, just like when using cargo test.
The <repofolder>/target/<config>/deps folder should be included in the LD_LIBRARY_PATH during test execution.

xd009642:
Should fix #1297


Luthaf:
Thanks for this! I'm happy to test it on the full project I have when the code is working on macOS.

xd009642:
@Luthaf tests pass on mac for latest build

Luthaf:
I can confirm that my code runs fine with this PR and fails on master!

xd009642:
Cool, I'm pretty happy with the impl it looks like it should be reliable enough. I'll merge this so it'll be available in the develop image and then tomorrow probably cut a new release

Issue #1621, opened 2024-09-05
Ptrace engine + no `--implicit-test-threads` causes passing quiche tests to fail


johnDeSilencio:
## Problem Statement

Tests in the [quiche](https://github.com/cloudflare/quiche/tree/master) crate that pass when running `cargo test` are failing when executed by `cargo tarpaulin` under the following conditions:

1. `ptrace` is the coverage tracing backend used
2. The flag `--implicit-test-threads` is omitted

## Expected Behavior

Tests that pass when executed by `cargo test` should also pass when executed by `cargo tarpaulin`

## Steps to Reproduce

1. Clone the [quiche repository](https://github.com/cloudflare/quiche)
2. Navigate to the `quiche/` subdirectory ([`quiche/quiche/`](https://github.com/cloudflare/quiche/tree/master/quiche))
3. Run the following command:

```none
cargo tarpaulin --engine Ptrace
```

The following tests will fail:

```none
failures:
    h3::tests::reset_finished_at_client
    h3::tests::reset_stream
    h3::tests::send_body_truncation_stream_blocked
    tests::collect_streams

test result: FAILED. 464 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 55.63s
```

I've attached the (compressed) [`ptrace` dump to this issue](https://github.com/user-attachments/files/16883965/tarpaulin_ptrace_dump.json.zip) as well as [the verbose `cargo tarpaulin` output](https://github.com/user-attachments/files/16884056/tarpaulin_output.txt).

## Workarounds

1. Using the `Llvm` engine instead of `ptrace`:

```none
cargo tarpaulin --engine Llvm

# ...

test tests::version_negotiation ... ok
test tests::zero_length_new_token ... ok
test tests::zero_rtt ... ok

test result: ok. 468 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 23.48s
```

2. Using the `--implicit-test-threads` flag:

```none
cargo tarpaulin --engine Ptrace --implicit-test-threads

# ...

test tests::version_negotiation ... ok
test tests::zero_length_new_token ... ok
test tests::zero_rtt ... ok

test result: ok. 468 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 43.08s
```

## Tooling & System Versions

`cargo-tarpaulin`:

```none
cargo-tarpaulin-tarpaulin 0.31.2
```

[...]

PR #1674, opened 2025-01-01
Move test threads to env var not CLI arg

xd009642:
#1621 and other past likely closed issues where people expected the args passed to the test binary were for their test and not to be consumed by cargo test should hopefully be fixed by this.


xd009642:
Well didn't fix what I hoped but still seems like a better idea

PR #1687, opened 2025-02-04
Handle doctests file paths that contain dashes

Nemo157:
When trying out the doctest support on a workspace involving many `kebab-case` named crates I noticed any `should_panic` doctests were failing.

I didn't manage to run the tests to confirm the added test reproduces it properly, I'm getting crashes inside the test runner on `develop` too. I'll spend a little more time tomorrow trying to diagnose what about my weird setup the test runner doesn't like.

### DIFF
--- a/src/cargo.rs
+++ b/src/cargo.rs
@@ -427,31 +427,26 @@ fn run_cargo(
 }
 
 fn convert_to_prefix(p: &Path) -> Option<String> {
-    // Need to go from directory after last one with Cargo.toml
-    let convert_name = |p: &Path| {
-        if let Some(s) = p.file_name() {
-            s.to_str().map(|x| x.replace(['.', '-'], "_")).unwrap_or_default()
-        } else {
-            String::new()
+    let mut buffer = vec![];
+    let mut p = Some(p);
+    while let Some(path_temp) = p {
+        // The only component of the path that should be lacking a filename is the final empty
+        // parent of a relative path, which we don't want to include anyway.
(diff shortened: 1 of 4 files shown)


Nemo157:
Managed to get the tests running, it looks like there is still an issue because `foo-bar/src/lib.rs` matches the `bar_src_lib_rs` binary too, I need to understand a bit better why it uses a suffix match.

xd009642:
So the doc test name creation sometimes changes in compiler versions and I've ended up through looking at diffs when they change it and a bit of trial and error getting something that mostly works. 

There is meant to be some metadata output in future that shows where they've been saved but I think it just needs someone to champion or push it forwards in the compiler - but as a short term thing potentially the rules I have can be tightened up now. Keeping an MSRV on a nightly kinda sucks so I'm happy to potentially break older nightlies within reason

Nemo157:
Right. So I guess the suffix matching is related to

https://github.com/xd009642/tarpaulin/blob/b95fd49c269660ed9c8e78612b5a17fbaa32efad/src/cargo.rs#L467-L469

Given that change should be over a year old at this point, could the matching be changed to be exact rather than suffix? Testing that it passes all the tests on a current-ish nightly.

xd009642:
Yeah I'm fine with an exact match, if you've done the change to test it mind PRing it? Otherwise I can do it later when I'm not meant to be doing the dayjob :tm:

xd009642:
I also fixed CI on dev if you want to rebase/merge in those changes

xd009642 (approved):

xd009642:
LGTM thanks for that :tada:

PR #1744, opened 2025-05-20
chore(docs): remove one-shot docker containers

reneleonhardt:
### Documentation
- `docker run` examples should remove one-shot containers to avoid manual work

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ jobs:
       - checkout
       - run:
           name: Coverage with docker
-          command: docker run --security-opt seccomp=unconfined -v "${PWD}:/volume" xd009642/tarpaulin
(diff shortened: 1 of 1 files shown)


reneleonhardt:
Hmm GitHub needs more AI I think, still not as smart as a real Dev... 😝
"No conflicts with base branch
Changes can be cleanly merged."
But all those changes except the duplicate linebreaks have already been applied, how can the diff still show "changes"? 🤣
https://github.com/xd009642/tarpaulin/pull/1744/files

xd009642:
Yeah github diffs can be a bit funny at times. Anyway I believe all these changes were already applied so I should be able to close this without merging (let me double check )

PR #1746, opened 2025-05-20
chore(docker): minimize docker context

reneleonhardt:
### Docker Build
- Minimize Docker context


reneleonhardt:
Why is the README explicitly included in the Docker context? 🤔

I don't know about tarpaulin, but isn't it possible to (pre) build it in the builder and just copy it into the final stage?
That would allow using the docker build cache to cache all dependencies between actual code rebuilds 😉

num_cpus doesn't seem to be used by the way...

xd009642:
That looks like a typo with the readme, it doesn't seem to be in the images at all.

Also num-cpus is used just only for the ptrace instrumentation (linux). I'll move that to be only under that config in my own update PR I'm doing now

xd009642:
Will do

PR #1766, opened 2025-08-13
Add new markdown report export

sneakycrow:
## Motivation
This PR adds Markdown as an export option. My team uses GitHub Actions exclusively for code coverage (without exporting to coveralls or another third party service). We wanted a way to show the coverage results in PR comments and GitHub Step Summaries.

### Context
Originally, I was just trying to parse the json output, but as far as I can tell I lose the context of the previous run in that scenario, which limits our ability to show the differences between two runs (as far as I can tell)

## Additional Notes
If desired, I can also update the README to show how to use GitHub Actions with Step Summary/PR replies for others that want to utilize that.

### DIFF
--- a/src/report/markdown.rs
+++ b/src/report/markdown.rs
@@ -87,6 +87,8 @@ pub fn export(coverage_data: &TraceMap, config: &Config) -> Result<(), RunError>
 
     // Next, add file-by-file coverage table
     markdown_content.push_str("## File Coverage\n\n");
+    // Wrap the table in a details element for collapsibility
+    markdown_content.push_str("<details>\n<summary>Show file coverage details</summary>\n\n");
     markdown_content.push_str("| File | Coverage | Percentage | Change |\n");
     markdown_content.push_str("|------|----------|------------|--------|\n");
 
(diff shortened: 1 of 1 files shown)


src/report/markdown.rs
```diff
@@ -0,0 +1,137 @@
...
+                    "-".to_string()
+                } else {
+                    let sign = if diff > 0.0 { "+" } else { "-" };
+                    format!("{}{:.2}%", sign, diff.abs())
+                }
+            }
+            None => "-".to_string(),
+        };
+
+        let path_str = path.to_string_lossy();
```
xd009642, line 124: So one consideration with this is that it's the absolute path of the file. You might want to mention the root directory somewhere else and then use a relative path instead (just to avoid one potentially very wide column)

sneakycrow: Rock on, done and done. Unsure if there's a better way to calculate that relative path, open to alternative more-cleanlier options

sneakycrow:
@xd009642 Added the comment suggestions. Also included a collapsible wrapper around the table itself for the _really long_ tables. Bit more of an opinionated update but works well in most Markdown contexts.

xd009642 (approved):
Cool LGTM. I'll try and do a release this week so you don't need to do a git install

PR #1779, opened 2025-10-04
feat(html): add dark theme

t-webber:
This PR adds a button to switch to dark theme. The light theme wasn't changed so the colours are the same in that case.

image

image

image

image


t-webber:
Great! I tried to guess the prettier configuration for it not to change too much of the file, please let me know if you prefer some other configuration settings

PR #1784, opened 2025-10-18
Fix a bunch of cargo config stuff...

xd009642:
So the previous way cargo-tarpaulin handled the config files was a bit lazy. It only looked in the project directory and the home directory. Now I follow the actual compiler behaviour and go from project to root looking for config files then the one in the home directory merging things as we go...

This also grabs the env vars from the config files and applies them when the metadata is gathered, project built and project ran!


xd009642:
So cargo config is hierarchical, it checks the current working directory then all the parent directories until root looking for a `.cargo/config` and then the the home directory. At each stage it also applies a merging of the configs (which matters for rust flags as it's a list). Plus there's the env table which is now applied. 

Ultimately I use cargo-config2 to do it now which replicates how cargo does it. Before I just checked current working directory and home, didn't do the merge and ignored the env section

PR #1794, opened 2025-11-18
fix #1793

### LINKED ISSUE #1793: SourceAnalysis::visit_mod needs to consider inline modules correctly
**Describe the bug**
If a module has an inline `cfg(test)` module, any module with that inline module's name is automatically ignored even if it is a "live" module due to the way `SourceAnalysis::visit_mod` updates `ctx.ignore_mods`.

**To Reproduce**

In `src/foo/blah.rs`:

```rust
#[cfg(test)]
mod over_here {
}
```

In `src/foo/over_here.rs`:

```rust
mod impl_over_here;
```

Because `src/foo/blah.rs` adds `src/foo/over_here` to the `ignore_mods`, `src/foo/over_here/impl_over_here.rs` is excluded. Spooky action at a distance :) .

**Expected behavior**
Inline modules should be more careful about adding paths to `ctx.ignore_mods`. I'm not sure of the 100% correct fix (I spent enough time chasing this down that I want to capture it here before thinking about solutions).

Of course, once this is fixed, `#[path]` just opens other cans of worms as well.

xd009642:
If we're ignoring an inline mod don't ignore the file. But it also, might not be enough... I'm not sure if this works with inline modules in `mod.rs` that are ignored and then have a file or if that's possible?


mathstuf (approved):
Looks good to me!

Issue #1791, opened 2025-11-07
Add line count to HTML output


TheJanzap:
**Describe the feature**
I would like to see the number of hits in the HTML report so I used
`cargo tarpaulin --out html --count`, as I thought that would add the number of hits to the lines, but it didn't. Would it be possible to add that?

(on a related note: What output formats are affected by `--count`, anyway?)

TheJanzap:
Works for me! Thanks for your MR!

PR #1801, opened 2025-12-02
#1791: Add line count to HTML output

alt440:
Removed tooltip for the covered line that was indicating the number of hits for the line.

Instead, replaced it by a small indicator at the right of the line number clearly indicating "Hit:<number>".

On hover, a tooltip appears that indicates the full number if it could not be fully visible from the limited space (useful when the value is above 999) and to provide more context


xd009642 (approved):
I like it, thanks :smile:

Issue #1676, opened 2025-01-03
Tracking coverage on crates outside the currently executing workspace


Nemo157:
In some situations it's necessary to have tests defined in a different workspace to the library. Currently afaict there is no way to record the coverage data from these tests.

Here's an example setup:

```
├── Cargo.toml
├── foo-tests/
│   ├── Cargo.toml
│   └── tests/
│       └── foo.rs
├── src/
│   └── lib.rs
└── tarpaulin.toml
```

```toml
# Cargo.toml
[package]
name = "foo"
version = "0.1.0"
edition = "2024"
```

```toml
# foo-tests/Cargo.toml
[package]
name = "foo-tests"
version = "0.1.0"
edition = "2024"

[dependencies]
foo.path = ".."
```

```rs
// foo-tests/tests/foo.rs
#[test]
fn test() {
  foo::add(1, 2);
}
```

```rs
// src/lib.rs
pub fn add(left: u64, right: u64) -> u64 {
    left + right
}
```

```toml
# tarpaulin.toml
[tests]
manifest-path = "foo-tests/Cargo.toml"
```

Running coverage on it gives:

```console
> cargo tarpaulin
...
2025-01-03T13:16:37.775101Z  INFO cargo_tarpaulin::process_handling: running target/x86_64-unknown-linux-gnu/debug/deps/foo-be23343dc3c96be8

running 1 test
test test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

2025-01-03T13:16:37.932096Z  INFO cargo_tarpaulin::report: Coverage Results:
|| Uncovered Lines:
|| Tested/Total Lines:
||
NaN% coverage, 0/0 lines covered
```

The `src/lib.rs` coverage data isn't recorded because it fails this check:

https://github.com/xd009642/tarpaulin/blob/c493bc1d9e581b0c4bca5e3c07868a5579286cbb/src/path_utils.rs#L64

I think what is needed is the ability to have the root for the recorded coverage different to the root of the workspace used for resolving relative paths. (In the real project I ran into this on we have tests in both the root workspace and multiple sub-workspaces that we want to merge all together into one coverage report, with the majority of library code in the root workspace).

xd009642:
A bit late to this but I was finishing off my holidays! I guess I'd always worked on the assumption that projects like that would make use of workspaces but this makes sense and is something tarpaulin should support. 

[...]

PR #1818, opened 2026-02-04
Coverage in workspace package

alt440:
#1676 

These changes separate root from manifest-path. 

Root is now set by default to the path 'cargo tarpaulin' is run from.

Root has been renamed to current_dir

### DIFF
--- a/tests/test_types.rs
+++ b/tests/test_types.rs
@@ -24,6 +24,7 @@ fn mix_test_types() {
     let test_dir = get_test_path("all_test_types");
     env::set_current_dir(&test_dir).unwrap();
     let mut manifest = test_dir.clone();
+    config.set_current_dir(test_dir.clone());
     manifest.push("Cargo.toml");
     config.set_manifest(manifest);
     let mut target = test_dir;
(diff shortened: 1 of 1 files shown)


xd009642:
I wanted to do some testing on this on other projects, and it doesn't seem to quite work correctly in reporting. To reproduce here's a small test project I did which you can compare how it runs with the current tarpaulin and this PR.

[test_project.zip](https://github.com/user-attachments/files/25989445/test_project.zip)

On this PR in the root directory if I do:

* `cargo tarpaulin --engine llvm --manifest-dir project/Cargo.toml`
* `cargo tarpaulin --engine llvm  --current-dir project/ --manifest-dir project/Cargo.toml`

I get no results printed out.

If I then cd into project and do

* `cargo tarpaulin --engine llvm` 

I still get no results printed out.

In contrast, with tarpaulin develop in the root when I do:

* `cargo tarpaulin --engine llvm --manifest-path project/Cargo.toml`

And in project do:

* `cargo tarpaulin --engine llvm` 

I get coverage results printed out.

tests/doc_coverage.rs
```diff
@@ -164,6 +170,7 @@ fn rustdocflags_handling() {
     let test_dir = get_test_path("rustflags");
     env::set_current_dir(&test_dir).unwrap();
     let mut manifest = test_dir;
+    config.set_current_dir(manifest.clone());
```
xd009642, line 173: A thought on this. Above there's `env::set_current_dir`. So this `set_current_dir` shouldn't be necessary.

However, it is worth doing a test where it's set via `env::set_current_dir` in the test and left default for `Config` and one where it's not set in the test and it's set via `Config` to make sure they work the same and there's not an issue in the `config.set_current_dir`.

PR #1831, opened 2026-03-22
Tests/simulation testing

xd009642:


xd009642:
Some notes for future me:

1. Better way to advance that process is to do it for every PID. Then things that are pending a continue etc don't do anything for but things that are "free running" move them on. This might need a different advance process function for non-continue advances :thinking:

PR #1834, opened 2026-03-25
Fix decode URL hash to support non-ASCII folder names

pengjiaxusz:
Apply decodeURIComponent to hash path components to fix navigation issues with Chinese and Unicode folder names.


xd009642 (approved):

xd009642:
xiexie :pray:

PR #1839, opened 2026-04-19
Fix sub-module path resolution for lib.rs/main.rs/mod.rs

### LINKED ISSUE #1837: Tarpaulin (or rather, LLVM?) misses #[cfg(test)] annotation and considers a test module as normal code
**Describe the bug**

I'm encountering a problem with cargo-tarpaulin 0.35.2 on macOS and Linux, where a module like this:

(from https://github.com/golddranks/kamekichi-bot/blob/main/crates/kamekichi-ws/src/lib.rs#L14)

```rust
#[cfg(test)]
mod tests;
```

...is considered as a code under test by cargo tarpaulin.

I didn't use to have this problem locally, with tarpaulin 0.34.1 for macOS, installed from Nix (25.11 stable).

However, on Linux CI (using LLVM), and locally after updating to 0.35.2, this problem started to manifest. Here is a log from the CI:

(from https://github.com/golddranks/kamekichi-bot/actions/runs/24284461162/job/70911440630#step:6:131)

```
2026-04-11T14:25:32.080967Z  INFO cargo_tarpaulin::report: Coverage Results:
|| Uncovered Lines:
|| crates/kamekichi-ws/src/lib.rs: 170, 212, 221, 268, 286, 301, 303, 305, 317, 533, 563-564, 842, 846, 858
|| crates/kamekichi-ws/src/read_buf.rs: 126
|| crates/kamekichi-ws/src/tests.rs: 80-81, 83-84, 587-589, 591-592, 1340-1341, 1349-1350, 1358-1359, 1367-1368, 1379-1380, 1385-1387, 1401-1402, 1407-1409, 1441-1442, 1660-1661
|| Tested/Total Lines:
|| crates/kamekichi-ws/src/error.rs: 74/74
|| crates/kamekichi-ws/src/lib.rs: 376/391
|| crates/kamekichi-ws/src/read_buf.rs: 61/62
|| crates/kamekichi-ws/src/rng.rs: 12/12
|| crates/kamekichi-ws/src/send_buf.rs: 38/38
|| crates/kamekichi-ws/src/tests.rs: 182/213
|| 
94.05% coverage, 743/790 lines covered
```

As you can see, it considers parts of `crates/kamekichi-ws/src/tests.rs` uncovered, although that is a test module.

**To Reproduce**

I don't think this is necessarily easy to reproduce, since it seems to depend on fiddly details such as tarpaulin version, LLVM version, rustc version. It currently reproduces both in Linux and macOS, with my kamekichi-ws repo, though.

**Expected behavior**

I expect `#[cfg(test)]` modules not to be included in the code coverage. However, I was unable to find in the README.md whether this is intenional or not.

golddranks:
Fixes https://github.com/xd009642/tarpaulin/issues/1837

Problem: In the source code analysis phase, `visit_mod` appended the enclosing file's stem to the child module path when determining where the module should live on disk, so `#[cfg(test)] mod tests;` in `lib.rs` resolved to `src/lib/tests.rs` instead of `src/tests.rs`. The bogus path never matched the source walker, so the child file escaped the ignore filter and its non-#[test] items were reported as uncovered, as reported in the issue above.

Fix: Skip the stem when the enclosing file is a crate root (lib.rs, main.rs) or directory module (mod.rs); those are transparent in Rust's module tree. Apparently, `ignore_nested_modules` already had the lib/mod guard but `visit_mod` did not. Btw. ignoring `main.rs` was missing from both, but I left `ignore_nested_modules` changes out of scope, as it wasn't relevant for this exact fix. A similar pattern was also found in `src/source_analysis/mod.rs:428` (Might be better to check if those warrant changes too?)

I also added some comments and renamed some local variables, to reflect my understanding of the code.

I think this targeted fix manages to fix the issue linked above, but I have some concerns about the codebase based on these short explorations: the module resolution is basically replicating what Cargo and rustc are doing already, and it would be much better to delegate those parts, if possible. I don't know how feasible that would be, though, so just a note.

Disclaimer: I used Claude Code Opus 4.7 to explore the causes of the bug and have suggestions for the fixes. The code in the actual fix in `src/source_analysis/items.rs` is fully mine. However, the test code in `src/source_analysis/tests.rs` is generated by Claude and reviewed by me.

### DIFF
--- a/src/source_analysis/items.rs
+++ b/src/source_analysis/items.rs
@@ -81,8 +81,8 @@ impl SourceAnalysis {
         let _guard = ctx.push_to_symbol_stack(module.ident.to_string());
         let analysis = self.get_line_analysis(ctx.file.to_path_buf());
         analysis.ignore_tokens(module.mod_token);
-        let check_insides = self.check_attr_list(&module.attrs, ctx);
-        if check_insides {
+        let should_cover = self.check_attr_list(&module.attrs, ctx);
+        if should_cover {
             if let Some((_, ref items)) = module.content {
                 self.process_items(items, ctx);
             }
@@ -94,14 +94,25 @@ impl SourceAnalysis {
                     self.ignore_nested_modules(items, ctx);
                 }
             } else {
-                // Get the file or directory name of the module
+                // This item was determined not to be covered, so we ignore it.
+                // Do a best-effort calculation of the path of the module
+                // corresponding to the mod item to ignore it.
+
+                // Get the file or directory name of the module containing the mod item
                 let mut p = if let Some(parent) = ctx.file.parent() {
                     parent.to_path_buf()
                 } else {
                     PathBuf::new()
                 };
-                if let Some(s) = ctx.file.file_stem().and_then(|x| x.to_str()) {
-                    p.push(s);
+                match ctx.file.file_stem().and_then(|x| x.to_str()) {
+                    Some(name) if ["lib", "mod", "main"].contains(&name) => {
+                        // skip because these segments are transparent in module tree
+                    }
+                    Some(name) => p.push(name),
+                    None => warn!(
+                        "Could not read file stem of {:?}; sub-module path may be wrong",
+                        ctx.file
+                    ),
                 }
                 let stack = ctx.symbol_stack.borrow();
                 for s in stack.iter().take(stack.len() - 1) {
--- a/src/source_analysis/tests.rs
+++ b/src/source_analysis/tests.rs
@@ -1844,6 +1844,63 @@ fn module_nesting_correct() {
         .ignore_mods
         .borrow()
         .contains(&PathBuf::from("src/foo/bar/inner.rs")));
+
+    // Top-level `#[cfg(test)] mod foo;` in lib.rs must resolve to src/foo.rs,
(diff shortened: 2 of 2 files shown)


xd009642:
LGTM

> I have some concerns about the codebase based on these short explorations: the module resolution is basically replicating what Cargo and rustc are doing already

Well parsing the source and creating this relationships is purely rustc (I don't think cargo does anything apart from consume and display the rustc error messages). And rustc isn't really library-fied so it's not like I can pull it in and get this info syn is how I have to live

golddranks:
Thanks! And yeah, I get it. Btw. how do you think about this? 

> I left ignore_nested_modules changes out of scope, as it wasn't relevant for this exact fix. A similar pattern was also found in src/source_analysis/mod.rs:428 (Might be better to check if those warrant changes too?)

xd009642:
Looking at that code I think it's fine, it's a mildly more tricky case where there's say a `src/foo.rs` that's ignored because it's a test module and there's a `src/foo/also_ignored.rs`. That match statement removes the `lib.rs` and `mod.rs` from the path to get the real module name/root. 

It is relatively new code iirc fixing another issue so there are a few tests to cover what it was fixing

PR #1841, opened 2026-04-20
Fix false-negative on inert match arms in generics

### LINKED ISSUE #1840: Inert match patterns in generic functions are showing up as uncovered even though they shouldn't
**Before raising the issue**

Using `--engine llvm` on macOS. Tarpaulin version 0.35.2, Rust 1.95.0.

**Describe the bug**

An should-be-covered (or should be ignored) region appears as noncovered whenever the following conditions hold:

- the containing function is generic
- there is a match with inert pattern (no match guard code or the actual match arm code on the same line)
- the value of the block is returned, not dropped in place

After digging a bit into it, it seems that:

- generic functions are "force-covered" by tarpaulin, and some parts that don't get LLVM instrumentation, are then ignored by pattern matching against the syntax
- LLVM doesn't emit instrumentation for lines where there is an inert match pattern
- UNLESS, the MIR block has a "drop" terminator.

This explains the bug. For a fix, it seems that this is another pattern where the line should be ignored by pattern matching. (A proper fix would be to actually not force-cover generics, but backpropagate LLVM's regions from the monomorphisated functions, but that seems like a huge amount of work)

**To Reproduce**

I have a repro here: https://github.com/golddranks/tarpaulin_test/blob/main/match_arm_problem/src/lib.rs

See this report: https://golddranks.github.io/tarpaulin_test/tarpaulin-report-macos-latest.html#match_arm_problem/src/lib.rs

**Expected behavior**
The match patterns that are actually covered, should be shown as covered, or possibly ignored.

golddranks:
Fixes https://github.com/xd009642/tarpaulin/issues/1840

Force-covered generic fns reported the `{` line as uncovered:

    pub fn f<S>() -> bool {
        match true {
            _ => {          // <-- reported uncovered
                true
            }
        }
    }

LLVM emits no coverage region for the brace line when the pattern is inert (no binding), so force-cover flags it as a miss. Ignore the brace line (through the first statement) when the pattern is inert (wildcard, literal, or-of-literals, tuple-struct of inert sub- patterns, ...), there's no guard, and no sibling arm shares the line. Binding patterns (`a =>`, `Some(x) =>`) and guarded arms are left alone — they already get coverage regions.

~~`Pat::Ident` is conservatively treated as a binding even for unit variants like `None` — syn can't tell them from fresh bindings without type resolution.~~ (This is now decided with a heuristic: SCREAMING_CASE and CamelCase are existing idents, whereas snake_case is a name binding.)

### DIFF
--- a/src/source_analysis/expressions.rs
+++ b/src/source_analysis/expressions.rs
@@ -134,7 +134,7 @@ impl SourceAnalysis {
     fn visit_match(&mut self, mat: &ExprMatch, ctx: &Context) -> SubResult {
         // a match with some arms is unreachable iff all its arms are unreachable
         let mut result = None;
-        for arm in &mat.arms {
+        for (arm_idx, arm) in mat.arms.iter().enumerate() {
             if self.check_attr_list(&arm.attrs, ctx) {
                 let reachable = self.process_expr(&arm.body, ctx);
                 if reachable.is_reachable() {
@@ -145,6 +145,7 @@ impl SourceAnalysis {
                     }
                     result = result.map(|x| x + reachable).or(Some(reachable));
                 }
+                self.maybe_ignore_inert_match_arm(arm, arm_idx, &mat.arms, ctx);
             } else {
                 let analysis = self.get_line_analysis(ctx.file.to_path_buf());
                 analysis.ignore_tokens(arm);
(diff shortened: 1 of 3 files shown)


golddranks:
The test ideas were basically sourced from [here](https://golddranks.github.io/tarpaulin_test/tarpaulin-report-macos-latest.html#match_arm_problem/src/lib.rs), but "positive" test cases were also added to make sure that lines in match patterns that shouldn't get ignored, don't accidentally get ignored.

golddranks:
I'm looking into one more last bug report I filed, the `loop` one, and it looks very similar, but that made me think that because the LLVM regions are working better on monomorphic code, it might be better to enable this pattern-matched ignoring behaviour only when working with generics, i.e. "force-covered" functions.

xd009642:
I'm just about to start work so will have to look more fully tomorrow, but one note on the existence of the source analysis module. It does exist because the original ptrace backend if things like generic functions aren't monomorphised they don't appear in debug information so we have to manually do some including. So some of the includes can be less eager for sure. 

But also it might be some of this code can be skipped for llvm coverage since it ensures now the counters remain in if the code isn't monomorphised. The early release of the llvm coverage in rustc the counters got removed via dead code elimination as well so this was still necessary. Though that's just a thought on improving the llvm results quality - it's still a good improvement for the ptrace backend :smile:

golddranks:
I'll mark this a draft, as I discovered more corner cases. I'll send the `loop` fix first, it seems to be the simpler one.

golddranks:
> But also it might be some of this code can be skipped for llvm coverage since it ensures now the counters remain in if the code isn't monomorphised. The early release of the llvm coverage in rustc the counters got removed via dead code elimination as well so this was still necessary. Though that's just a thought on improving the llvm results quality - it's still a good improvement for the ptrace backend 😄

I'm not sure if I understand this. Does that mean that the _current_ llvm would support "backpropagating" the coverage from the monomorphised instances to the generic one, and that didn't use to be the case?

xd009642:
> I'm not sure if I understand this. Does that mean that the current llvm would support "backpropagating" the coverage from the monomorphised instances to the generic one, and that didn't use to be the case?

So the way it used to work, if the code wasn't generated via monomorphisation then there were no counters and the lines wouldn't show up as coverable. Now there's like stub counters that are in the program headers for coverage info so if it's never monomorphised it shows up in coverage. 

FWIW I think this change actually went in before the llvm coverage was on stable rust, it's just tarpaulin is older than that by a good few years.

golddranks:
Okay, I set this as ready for review now. The second commit might be a bit controversial, as it uses heuristic for deciding if something is a binding or a path, so I split it that it can be easily dropped if you are against it. I believe the heuristic is reasonable enough, though.

xd009642 (approved):
I'm personally fine with the heuristic, LGTM thanks 🎉

